Issue #3, opened 2016-11-30
R&D porting the existing pilot codes to add A-Z links to Facet view pages


lsitu:
This will include the following feature in dams5-cc-pilot:
https://github.com/ucsdlib/dams5-cc-pilot/issues/33

hweng:
Porting code for this feature has been implemented in https://github.com/ucsdlib/horton/pull/5

mcritchlow:
@hweng this work was completed, right?

hweng:
@mcritchlow  I think most is done. Maybe just need to update the catalog_controller configuration to upper-case letter for  A-Z links.

mcritchlow:
OK, so enhancing what you've done in this ticket already. I'll set that up as a new ticket for next Sprint. If there are other related blacklight things we should port, please let me know.

PR #5, opened 2016-11-30
#3 Close -Adding facet browsing page and a-z links.

hweng:


hweng:

lsitu:
@hweng What's the expected behavior if I have Topics like "ATopic", "BTopic", "NTopic" etc. when I click on letter 'B' and letter 'N' from http://localhost:3000/catalog/facet/topic_sim?facet.prefix=s&facet.sort=index? I got nothing when I click on those letters while these facet subjects are displayed on the index page. Will listing by letters be triggered when more topics added?

hweng:
@lsitu The blacklight built-in function only support for either Upper case or lower case, but not both. But since the solr facets sorting will use LowerCaseFilterFactory to make it case insensitive https://github.com/ucsdlib/dams5-cc-pilot/issues/32, which all of the facets would convert to low case so the above problem will not happen.  I will create the ticket to port solr sorting code to this project.

lsitu:
@hweng It looks like a different issue. This are the examples which could be something like "Aopic", "Bopic", "Nopic". But when I click on the letters like 'A', 'B', or 'N', I see no results displayed. Do you see the same result? We should expect 'Aopic' if I click on letter 'A', and 'Bopic' for letter 'B', right?

hweng:
@lsitu It is the same issue I mentioned above. Can you create a topic start with lower case such as "apple", "banana" and etc, and remember to update your object for the topic? Then go to browse by topic a-z link page, it should appear when you click "a" or "b.

hweng:
For example, here is the result when I click "t". But now you need to create the one start with lowercase . And Later after I port solr sort config code, it will both work whether you create lower case or upper case topics.  Hope it helpful!

screen shot 2016-11-30 at 5 09 21 pm

lsitu:
@hweng Yes, I see that, which is case sensitive. It looks like the default should be uppercase since the label of an authority is stating with uppercase in general.

hweng:
@lsitu If change to uppercase, it won't work after port solr sort config. We don't need to worry about record label, because it, such as label of an authority is different set from facets in solr.

lsitu:
@hweng Thanks for clarifing that. Is it possible to add a test case for those letter links to make sure the facets do showing up when clicking on it?

hweng:
@lsitu Yes, it was added already:
https://github.com/ucsdlib/horton/blob/2e337e6665a7bad1469f06a0057867a6a2666c0b/spec/features/catalog_spec.rb

VivianChu:
👍

lsitu:
@hweng I mean a test to demonstrate for a facet topic like 'btopic' to be showing up when clicking on the link for the letter 'b'.

hweng:
@lsitu It is built-in function in blacklight which should be tested in blacklight. Do we need to do this redundant test here in local app? I think what we need to make sure that this feature has been enabled to two facets "topic" and "creator", which is demonstrated in the test already.

lsitu:
@hweng I think that's fine and I'll merge it now. Just feeling so bad that I have to test it manually and see nothing is showing up.

hweng:
@lsitu All facets start with lower case should work. And after i port the solr config to make the facets case insensitive, your issue will be resolved automatically.
And also I will post the issue that only support either lower or upper case but not both to blacklight channel and see how other folks think.

lsitu:
@hweng It sounds good. Thanks.

PR #6, opened 2016-12-05
Updated solr config for facet sorting- dams5-cc-pilot#32

hweng:
This issue was raised in the Review meeting from Sprint 21. Questions:

Can lexical sorting be used in a way similar to the description in this article?
Can we ignore case? Example of ALLCAPS values
Can we ignore punctuation?

### DIFF
--- a/solr/config/schema.xml
+++ b/solr/config/schema.xml
@@ -125,6 +125,10 @@
     <fieldtype name="alphaSort" class="solr.TextField" sortMissingLast="true" omitNorms="true">
       <analyzer>
         <tokenizer class="solr.KeywordTokenizerFactory" />
+        <filter class="solr.LowerCaseFilterFactory"/>
+        <filter class="solr.PatternReplaceFilterFactory" pattern="([^A-Za-z0-9])" replacement="" replace="all"/>
+        <filter class="solr.PatternReplaceFilterFactory" pattern="(\d+)" replacement="000$1" replace="all"/>
+        <filter class="solr.PatternReplaceFilterFactory" pattern="0*([0-9]{6,})" replacement="$1" replace="all"/>
         <filter class="solr.ICUFoldingFilterFactory"/>
         <filter class="solr.TrimFilterFactory" />
       </analyzer>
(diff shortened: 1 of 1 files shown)


hweng:

Note: You existing records in your local environment need to reindex to get it work. Before re-index your local records, make sure the solr schema in your local development has been up-to-date by checking here: http://localhost:8983/solr/#/hydra-development/files?file=schema.xml

hweng:
@lsitu These are the solr filters which should be tested well by solr communities. And also I consider  some potential issues for adding tests for solr config changes, because people need to manually update their local test environment to add changes to solr config so that the test will pass on their local.

lsitu:
@hweng I think we may need to prove the filters you are using works as expected for us. We can set up our local environments to make it all the tests passed as long as your tests can be passed locally. What do you think?

hweng:
@lsitu, And for your 2nd question. Did you update your development environment solr schema which located in the tmp directory, and reindexed your records?

lsitu:
@hweng I think we may just need to prove those features we added work as expected and it won't break other existing features. Manually test everything may not be the way we like unless automatic test is not possible for that feature. I believe horton is in a better position than dams5-cc-pilot now so we may want some basic automatic tests to prove that the features we added will work as expected.

hweng:
@lsitu The expected solr filter behavior is documented in the solr documents.  The steps we need to do now is:
1. Show the end-results to Gabriela or Awren and etc, which I will make demo this Friday.
2. Get feedback from them if we want to use it or not.
3.  If we decide to use these filters, i will consider to add tests of its behavior even though I have concern about whether it is good or not to add such tests.

lsitu:
@hweng Okay. Then I'll wait.

hweng:
@lsitu For your previous comment, I meant your development environment. If you don't update your development environment, you will not get the results as expected which stated in solr document.

VivianChu:
👍

lsitu:
@hweng Is it a build-in command that we can use to re-index all the records as you noted in your [comment](https://github.com/ucsdlib/horton/pull/6#issuecomment-264956886) above?

hweng:
@lsitu I mentioned above is just to remind you that you need to make sure your solr schema.xml up-to-date in your local development and re-index all records.
And as you know that for any solr config changes you need to do that.

hweng:
The group decided not to implement the above solr filters to facet ordering now.

lsitu:
@hweng Thanks for clarifying that. Yes, we do need to find a way to re-index all records.

PR #7, opened 2016-12-07
Fixes - implementing group access controls. dams5-cc-pilot#44

lsitu:
https://github.com/ucsdlib/dams5-cc-pilot/issues/44


lsitu:
@ucsdlib/developers Please review, comments ... etc.

VivianChu:
👍

hweng:
@lsitu It looks fine to me, except that I see there are several classes copied over from curation_concerns to our local app with only one line added, for example:
app/search_builders/curation_concerns/embargo_search_builder.rb
app/search_builders/curation_concerns/file_set_search_builder.rb
app/search_builders/curation_concerns/lease_search_builder.rb
.....

Would it cause some future issues like Upgrading the curation_concerns gem causing the local app broken?  Is there better way.  Otherwise :+1:

lsitu:
@hweng Yes, it could be but it should be an easy fix as well since the change is wrapped up in one line. With the merging of CC/Sufia is on going, I would expect some big changes in the future.

lsitu:
@ucsdlib/developers Could anyone convenient to merge it for now? We may want other features to build on top of this group access controls. Thanks.

lsitu:
@ucsdlib/developers Thank you all for the reviewing and merging it. Yes, we could look forward to the new CC/Sufia and see what's the better way for it. But now, I am just trying to make it easy for us to track and update later.

PR #9, opened 2016-12-19
Fixes adding encrypted streaming support for A/V files via Wowza #43

lsitu:
Related https://github.com/ucsdlib/dams5-cc-pilot/issues/43, which is pair with https://github.com/ucsdlib/encrypted-stream-name/issues/1.

### DIFF
--- a/app/helpers/wowza_helper.rb
+++ b/app/helpers/wowza_helper.rb
@@ -1,0 +1,44 @@
+module WowzaHelper
+  #---
+  # Builds Wowza URL
+  #
+  # @param obj_id
+  # @param file_set_id
+  # @param fedora_file
+  # @media_type: 'audio', 'video'
+  # @return string or nil
+  #---
+
+  def wowza_url(obj_id, file_set_id, fedora_file, media_type, request_ip)
+    plain_text = "#{obj_id} #{file_set_id} #{fedora_file} #{media_type} #{request_ip}"
+    encrypted = encrypt_stream_name plain_text
+    return Rails.configuration.wowza_baseurl + encrypted
+  end
+
+
+  ## video stream name encryption
+  def encrypt_stream_name( plaintext )
+
+    # random nonce
+    nonce=rand(36**16).to_s(36)
+    while nonce.length < 16 do
+      nonce += "x"
+    end
+
+    # load key from environment variable
+    key = ENV.fetch('APPS_DHH_STREAMING_KEY')
+
+    # encrypt
+    cipher = OpenSSL::Cipher::AES.new(128,:CBC)
+    cipher.encrypt
+    cipher.key = key
+    cipher.iv = nonce
+    enc = cipher.update(plaintext) + cipher.final
+
+    # base64-encode
+    b64 = Base64.encode64 enc
+    b64 = b64.gsub("+","-").gsub("/","_").gsub("\n","")
+    "#{nonce},#{b64}"
+  end
+
+end
--- a/app/views/curation_concerns/file_sets/media_display/_audio.html.erb
+++ b/app/views/curation_concerns/file_sets/media_display/_audio.html.erb
@@ -1,0 +1,39 @@
+<% if Rails.configuration.wowza_enabled
+  solr_document = file_set
+  solr_document = file_set.solr_document if (file_set.respond_to? :solr_document)
+
+  parent_id = parent.id if defined?(parent) && !parent.nil?
+  fedora_file = solr_document['digest_ssim'].first.split(':')[-1] if !solr_document['digest_ssim'].nil?
+
+  file_ext = 'mp3'
+  file_ext = solr_document['label_tesim'].first.split('.')[-1] if solr_document['label_tesim'].first
+
+  wowzaURL = wowza_url(parent_id, file_set.id, fedora_file, file_ext, request.ip)
+%>
+  <% if wowzaURL != nil %>
+    <div id="dams-audio">Loading the player...</div>
+
+    <script type="text/javascript">
+      jwplayer("dams-audio").setup({
+        playlist:
+          [{
+            sources:
+              [
+                {file: "http://<%=wowzaURL%>/playlist.m3u8"}
+              ]
+           }],
+        width: "100%",
+        height: 60,
+        rtmp: {bufferlength: 3},
+        analytics: {enabled: false}
+      });
+    </script>
+  <% end %>
+<% else %>
+  <audio controls="controls" class="audiojs" preload="auto">
+    <source src="<%= main_app.download_path(file_set, file: 'ogg') %>" type="audio/ogg" />
+    <source src="<%= main_app.download_path(file_set, file: 'mp3') %>" type="audio/mpeg" />
+    Your browser does not support the audio tag.
+  </audio>
+<% end %>
+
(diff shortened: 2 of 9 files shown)


lsitu:
@ucsdlib/developers Please review, comments ... Thanks

mcritchlow:
@lsitu - there appears to be some duplicate commits in this PR and #12 Per my comments in #12. If you can consolidate the scope of the commits in that PR to non A/V files and leave this PR to A/V only, it might make things a bit easier to review and discuss.

lsitu:
@mcritchlow I think PR https://github.com/ucsdlib/horton/pull/9 is just overlapping with PR https://github.com/ucsdlib/horton/pull/12 for the first commit. While PR #9 is good for test the Wowza support by ingesting a MP4 video, we can discuss and look for a way to separate the support for A/V files and images in ticket https://github.com/ucsdlib/horton/issues/11 for PR #12.

mcritchlow:
@lsitu - ok thanks for clarifying. Can we separate the PR commits so there isn't any overlap?

lsitu:
@mcritchlow I think I can close PR #12 for now until PR #9 is merged, otherwise I have to move the codes in the second commit to a new branch, which need to copy and paste all the changes over there.

mcritchlow:
@lsitu - OK, sounds good. Should have time to look at #9 later today. And hopefully others can as well.

app/helpers/wowza_helper.rb
```diff
@@ -0,0 +1,44 @@
...
+  # Builds Wowza URL
+  #
+  # @param obj_id
+  # @param file_set_id
+  # @param fedora_file
+  # @media_type: 'audio', 'video'
+  # @return string or nil
+  #---
+
+  def grabWowzaURL(obj_id, file_set_id, fedora_file, media_type, request_ip)
```
mcritchlow: I know this is pulling over existing code, but we should probably clean it up a bit since we have the chance.

This method name should be changed to use the ruby method convention. And grab isn't particularly valuable. It could simply be `wowza_url` or something like `encrypted_streaming_url`, or something like that.

app/helpers/wowza_helper.rb
```diff
@@ -0,0 +1,44 @@
...
+  def encrypt_stream_name( obj_id, file_set_id, fedora_file, media_type, ip )
+
+    # random nonce
+    nonce=rand(36**16).to_s(36)
+    while nonce.length < 16 do
+      nonce += "x"
+    end
+
+    # load key from environment variable
+    key = ENV.fetch('APPS_DHH_STREAMING_KEY') {'xxxxxxxxxxxxxxxx'}
```
mcritchlow: What do we want to happen if the streaming key doesn't exist? I suspect we don't want to return this bogus `'xx...'` value. We could not have a fallback, in which case I believe fetch with return a `KeyError`. That seems reasonable to me, since the key should exist. But curious what others think.

app/views/curation_concerns/file_sets/media_display/_audio.html.erb
```diff
@@ -0,0 +1,40 @@
...
+  solr_document = file_set
+  solr_document = file_set.solr_document if (file_set.respond_to? :solr_document)
+
+  parent_id = parent.id if defined?(parent) && !parent.nil?
+  fedora_file = solr_document['digest_ssim'].first.split(':')[-1] if !solr_document['digest_ssim'].nil?
+
+  file_ext = 'mp3'
+  file_ext = solr_document['label_tesim'].first.split('.')[-1] if solr_document['label_tesim'].first
+
+  wowzaURL = grabWowzaURL(parent_id, file_set.id, fedora_file, file_ext, '127.0.0.1')
```
mcritchlow: Should the erb template be passing in (and therefore know about) a default value like the localhost URL? It seems like the method itself should take care of this sort of thing.

app/views/curation_concerns/file_sets/media_display/_audio.html.erb
```diff
@@ -0,0 +1,40 @@
...
+puts "#{parent_id} #{file_set.id} #{fedora_file} #{file_ext} 127.0.0.1: #{wowzaURL}"
```
mcritchlow: should this be a call to the logger instead of puts?

app/views/curation_concerns/file_sets/media_display/_audio.html.erb
```diff
@@ -0,0 +1,40 @@
...
+%>
+  <% if wowzaURL != nil %>
+	<div id="dams-audio">Loading the player...</div>
+
+	<script type="text/javascript">
+		jwplayer("dams-audio").setup({
```
mcritchlow: can we clean up the spacing here? Also the configuration may need to be updated based on the issues OPS noted in https://github.com/ucsdlib/damspas/issues/271. Depending on when @VivianChu works on that ticket, that could either get folded into this or as a separate ticket. But I want to make sure we don't lose those changes.

app/views/curation_concerns/file_sets/media_display/_video.html.erb
```diff
@@ -0,0 +1,42 @@
+<% if Rails.configuration.wowza_enabled
```
mcritchlow: same comments/questions for this ERB as the audio one.

app/helpers/wowza_helper.rb
```diff
@@ -0,0 +1,44 @@
...
+    
+    # encrypt
+    str="#{obj_id} #{file_set_id} #{fedora_file} #{media_type} #{ip}"
```
mcritchlow: This is just an idea to consider. But as far as I can tell, it seems that all of the arguments passed to this method are being used only once (here) to create the string to pass to the cipher to encrypt. Does it make sense to consolidate them into a single string to pass into this method as a single argument?

app/helpers/wowza_helper.rb
lsitu: Yes, both encrypted_streaming_url and wowza_url sounds good and I'll change it.

app/views/curation_concerns/file_sets/media_display/_audio.html.erb
lsitu: Good catch! It should be request.ip.

app/views/curation_concerns/file_sets/media_display/_audio.html.erb
lsitu: Yes, we can use Rails.logger.info. But I think it's something I need to clean up unless we really want to log it here. What do you think?

mcritchlow: Oh, yes if it's not needed I'd say just delete it 👍

lsitu:
@mcritchlow I've added commit https://github.com/ucsdlib/horton/pull/9/commits/84d9efe3b74d3a17dd0b7bcd0441e274893db425 to address the issues you raised, but leaving it open for the default 'xxx...' key. However, I am fine with raising an exception when the key is missing.

lsitu:
@mcritchlow I've updated my last commit to include throwing the KeyError when the streaming key isn't set with tests (see https://github.com/ucsdlib/horton/pull/9/commits/ee6c2c76d925a2ab6c7818fef6ecbc436800a0ef)

mcritchlow:
@lsitu - This looks great to me. Will let @hweng and/or @VivianChu review/merge

VivianChu:
👍

hweng:
@lsitu Looks :+: to me. Would you merge two commits into one. Then I will merge your pull request.

lsitu:
@hweng Those two commits are squashed into https://github.com/ucsdlib/horton/pull/9/commits/28f9237890c48b06d556797c491a8fed3125699d.

lsitu:
@ucsdlib/developers Thank you all for reviewing, commenting and merging the PR.

PR #10, opened 2016-12-21
Closed adding initial support for CSV batch ingest. dams5-cc-pilot #47

lsitu:
Related to https://github.com/ucsdlib/dams5-cc-pilot/issues/47


lsitu:
@ucsdlib/developers Please review, comments ... Thanks

app/controllers/concerns/csv_imports_controller_behavior.rb
```diff
@@ -0,0 +1,63 @@
...
+          redirect_to sufia.dashboard_works_path
+        end
+      end
+
+      def create_update_job
+        log = Sufia::BatchCreateOperation.create!(user: current_user,
+                                           operation_type: "CSV Import")
+
+        uploaded_files = params.fetch(:uploaded_files, [])
+        selected_files = params.fetch(:selected_files, {}).values
```
mcritchlow: If `uploaded_files` or `selected_files` hit the fallback case for `fetch` (returning [] or {}), then should we use that as a guard class and return? It looks like it's calling the `perform_later` method on the job regardless of there being files or not. Or, are files optional and this is desired behavior?

app/jobs/csv_import_job.rb
```diff
@@ -0,0 +1,112 @@
...
+        f = Sufia::UploadedFile.find(id).file if !id.blank?
+        uploaded_files_map[File.basename(f.current_path)] = id
+      end
+
+      selected_files.each do |f|
+        remote_files_map[f[:file_name]] = f
+      end
+
+      model = model_to_create(attributes) #::ObjectResource #
+      count = 0
```
mcritchlow: what is the purpose of `count`? I only see it being incremented at the end of the each block, but not used.

app/jobs/csv_import_job.rb
```diff
@@ -0,0 +1,112 @@
...
+      parser = CSVParser.new(csv_source)
+      parser.each do |attrs|
+        splict_attributes (attrs)
```
mcritchlow: is this supposed to be `splice_attributes`?

app/jobs/csv_import_job.rb
```diff
@@ -0,0 +1,112 @@
...
+        auth_arr = attrs.delete(:Language)
+        attrs = attrs.merge(language: auth_arr)
+
+        # Resource Type field
+        auth_arr = attrs.delete(:Type)
+        attrs = attrs.merge(resource_type: auth_arr)
+
+        matched_files = []
+        remote_files = []
+        if attrs.include?(:file_1)
```
mcritchlow: `:file_1` and `:file_2` seem like fairly arbitrary names. Are there more meaningful names we could use? Is one referring to a master file, for example?

app/jobs/csv_import_job.rb
```diff
@@ -0,0 +1,112 @@
...
+        matched_files = []
+        remote_files = []
+        if attrs.include?(:file_1) 
+          file_name = attrs[:file_1].first
+          matched_files << uploaded_files_map[attrs[:file_1].first] if uploaded_files_map.key?(file_name)
+          remote_files << remote_files_map[attrs[:file_1].first] if remote_files_map.key?(file_name)
+          attrs.delete(:file_1)
+        end
+
+        if attrs.include?(:file_2)
```
mcritchlow: These two if blocks, for `:file_1` and `:file_2` are exactly the same. Perhaps there's a method to extract here?

app/jobs/csv_import_job.rb
```diff
@@ -0,0 +1,112 @@
...
+
+    # Override this method if you have a different rubric for choosing the model
+    # @param [Hash] attributes
+    # @return String the model to create
+    def model_to_create(attributes)
+      Sufia.config.model_to_create.call(attributes)
+    end
+
+    def authority_hash (attrs, key, model)
+      authority_arr = []
```
mcritchlow: you might be able to use either `map` directly on `attrs` or `each_with_object` here to build the authority array.

app/jobs/csv_import_job.rb
```diff
@@ -0,0 +1,112 @@
...
+      if(attrs.key?(key))
+        attrs.delete(key).each do |label|
+          authority_arr << AuthoritiesService.find_or_create(model, label)
+        end
+      end
+      authority_arr
+    end
+
+    def splict_attributes (attrs)
```
mcritchlow: same name question, should this be `splice_attributes`?

app/controllers/concerns/csv_imports_controller_behavior.rb
lsitu: @mcritchlow Yes, the source files are optional for our ingest and we can choose from either uploaded_files (from local disk) or selected_files (from server side or other share drives), or even both as what've configured, and [] or {} are guard classes when calling the the perform_later method on the job.

app/jobs/csv_import_job.rb
lsitu: Yes, the counter hasn't been used yet and we can remove it for now.

app/jobs/csv_import_job.rb
lsitu: Yes, it's a typo for split_attributes, and splice_attributes makes sense too.

mcritchlow: I see. Yep, split probably makes the most sense. 👍

app/jobs/csv_import_job.rb
lsitu: Yes, `:file_1` and and `:file_2` are arbitrary names from the CSV template header (borrowed from the dams4 Excel Input Stream) and we only can change it when the new CSV template is changed.

app/jobs/csv_import_job.rb
lsitu: Sure. We can extract it out.

lsitu:
@mcritchlow I've added commit https://github.com/ucsdlib/horton/pull/10/commits/db3662735228ddf6557949ae3a809bf2b5f100de to address the issues you suggested above. Please let me know if I miss anything. Thanks.

hweng:
:+1:

VivianChu:
👍

Issue #11, opened 2016-12-22
R&D Fedora server side derivatives storage strategy


lsitu:
CC/Sufia stores all thumbnail/derivatives on local disk. This ticket will discover the potential support with Fedora server side thumbnails/derivatives storage.

Related issues: https://github.com/ucsdlib/encrypted-stream-name/issues/1, https://github.com/ucsdlib/dams5-cc-pilot/issues/43

lsitu:
@mcritchlow I've created a [PR](https://github.com/ucsdlib/horton/pull/12) that persists thumbnail/derivatives to Fedora instead of local disk storage that CC/Sufia persists thumbnails. We've got a thumbnail, a preview file, and a service file persisted along with the master file on Fedora.

mcritchlow:
@lsitu - possibly of interest, there was a refactoring of some of the code this depends on to extract out the derivative handling to a [service class](https://github.com/projecthydra-labs/hyrax/pull/288/commits/a38179c7840550770d565a6d57cc206fc1a33e8b).

mcritchlow:
@lsitu i'm curious what your thoughts are on moving ahead with this. Should we wait until we have a viable test environment to do performance testing? It seems like we might have a related dependency on IIIF implementation as well, at least for images. If so, we could perhaps have you dig into IIIF next Sprint.

lsitu:
@mcritchlow I believe it's still hard to make a final call for derivatives storage and potentially the master file storage. In Fedora 3, files are separated from Fedora, which is more convenient for migration and management. And now lots of institutions are looking for solutions to separate the files from Fedora. Maybe we can pay more attention to this approach too.
For IIIF, yes, it seems like a good option for us and I think we can implement it next. Thanks.

mcritchlow:
After speaking with @lsitu, @rstanonik and @jhriv and Mike last week, and again with @lsitu today, we've decided to move forward with the following:

* We will use the default Fedora filesystem persistence strategy
* We will use the default Hyrax derivatives persistence strategy. Master/source files to go Fedora, derivatives are put on disk. This will require a share/directory agreed upon with OPS.

The basis for these decisions is to try and follow community conventions as much as possible. If we run into issues with these conventions in the future, we can re-evaluate each choice. But this sets us up well for easier setup, configuration, and upgrades over time.

Derivative presentation may ultimately end up leading us towards a caching layer with something like Varnish, memcache, etc. But we can cross that bridge when we get to it (likely with IIIF).

Issue #13, opened 2017-01-04
Add Continuous Integration and related plugins/services


mcritchlow:
Please setup CircleCI, Code Climate, and access-lint accessibility integration for this repository.

PR #18, opened 2017-01-06
Fixes adding support to persist audio/video derivatives on Fedora. #11

lsitu:
Related: https://github.com/ucsdlib/horton/issues/11


lsitu:

app/indexers/file_set_indexer.rb
```diff
@@ -0,0 +1,13 @@
+class FileSetIndexer < CurationConcerns::FileSetIndexer
+  def generate_solr_document
+    super.tap do |solr_doc|
+      solr_doc[Solrizer.solr_name('service_file_digest', :symbol)] = digest_from_service_file_content
```
lsitu: I see it won't index a nil value field but yes I can add condition check for it.

spec/controllers/downloads_controller_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    context "when the user has access" do
+      before { sign_in user }
+
+      it 'sends the original file' do
+        get :show, :id => file_set.id
+        expect(response.body).to eq file_set.original_file.content
+      end
+
+      it 'sends the derivative' do
```
lsitu: I think this could be an issue since Sufia/CC don't have the service files and allow public users to download the original files. Could we set up another ticket for this matter as global support? We may need to determine what kind of intermediate files we need to have in DAMS5 first.

spec/models/file_set_spec.rb
```diff
@@ -1,7 +1,30 @@
...
+      before do
+        Hydra::Works::AddFileToFileSet.call(file_set, File.open("#{fixture_path}/files/video.mp4", 'rb'), :original_file, update_existing: true)
+      end
+      it 'makes a service file' do
+        file_set.create_service_file
+        expect(file_set.service_file).not_to be nil
+        expect(file_set.service_file.uri.to_s).to include(ActiveFedora.fedora.host)
+      end
+    end
+    describe 'with an audio', if: CurationConcerns.config.enable_ffmpeg do
```
lsitu: Parameter CurationConcerns.config.enable_ffmpeg is configurable and the default is false in Sufia/CC so I think it's not bad to add the condition there. But yes we can just remove it for our case CurationConcerns.config.enable_ffmpeg.

app/indexers/file_set_indexer.rb
mcritchlow: Ah, if it won't index a nil field then nevermind. Even better :)

spec/controllers/downloads_controller_spec.rb
mcritchlow: Yes, that's a great idea. I suspect this is relevant to what that FileSets working group is tackling. So if @remerjohnson joins it, or one of us, we can track whether upstream support might be coming for service files, etc.

spec/models/file_set_spec.rb
mcritchlow: Sounds good. Since we know we'll have it set to true in our app (as you did in this PR, seems like we can safely remove it.

lsitu:
@mcritchlow I've removed the conditional check for the parameter CurationConcerns.config.enable_ffmpeg and rebase my branch with master. However, ci/circleci failed for all tests. I believe we still need to set up the test environment with SOLR Wrapper, Fedora wrapper, redis-server etc.

spec/controllers/downloads_controller_spec.rb
lsitu: @mcritchlow Yep, it looks like the FileSets working group is trying to provide support to ingest multiple files to the Fileset with different use in the UI level as well. We will know what we can have from upstream soon and see what we need to do next.

lsitu:
@VivianChu It looks like the test environment failed to start the test servers with `rake hydra:test_server` and `redis-server` and all the tests are failed. Could you take a look? Thanks.

mcritchlow:
Despite the CircleCI issues (ongoing work in #13) I'm going to merge this so @lsitu can get started on #21

PR #22, opened 2017-01-10
Fixes upgrading to rails 5. #21

lsitu:
Related: https://github.com/ucsdlib/horton/issues/21


lsitu:
@mcritchlow I've upgraded it to Rails 5.0.0.1. However, there are tons of deprecation warnings, which are from the dependency gems:
[DEPRECATION] RDF::Queryable::Enumerator#to_ary is deprecated, use RDF::Queryable::Enumerator#to_a instead. Called from /Users/lsitu/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rdf-2.2.1/lib/rdf/mixin/enumerable.rb:125
[DEPRECATION] RDF::Enumerable::Enumerator#to_ary is deprecated, use RDF::Enumerable::Enumerator#to_a instead. Called from /Users/lsitu/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rdf-2.2.1/lib/rdf/mixin/enumerable.rb:444

I am not sure which gem use the deprecated methods. Should we move ahead with hyrax migration in this case?

mcritchlow:
@lsitu - Are the deprecation warnings all from the RDF gem? The deprecation warnings make sense, as the most [recent version](https://github.com/ruby-rdf/rdf/releases/tag/2.2.0) notes the changes you're seeing. I assume these are coming from either ActiveFedora (and)or Curation Concerns. Are there others you're seeing that we should look at as well?

lsitu:
@mcritchlow Yes, the deprecation warnings are from the the RDF latest gem. In addition to ActiveFedora (and)or Curation Concerns, I think we may keep an eye on ActiveTriples as well.

lsitu:
@mcritchlow I see an weird issue with CSV ingest with the rails 5.0.0.1 migration, files selected from browse-everything failed to be ingested but it works with files selected from local disk. Maybe we should upgrade it to the latest CC to see how it goes?

mcritchlow:
@lsitu - hmm, there was a comment about something similar in the hydra slack channel this morning, but not sure if it's directly related. In any event, I think trying out the upgrade of CC makes sense. If you run into further (or deeper) issues, we may want to circle back on this approach vs a clean start?

lsitu:
@mcritchlow It looks like there is an issue with rails 5 that preventing the remote files from being ingested successfully since the local files ingest looks good and the metadata records get ingested as well. So it should be something related to Sufia/CC. I've get something working with Hyrax migration now but still lots of uncertain issues that need to look back and forth for the changes in Hyrax to make it work. Hopefully we can call for the decision soon.

Issue #17, opened 2017-01-06
R&D: Queue options


mcritchlow:
As Hyrax uses `active_job` to support message queues, we have a few options to consider in implementing our own queue system. The two obvious candidates are:

* Resque
* Sidekiq

There is a mix of usage in the community, so we should probably do a little research and ask around for people's experience, particularly as it relates to larger batch processes (since that will be our primary usage of the queuing system). I believe Princeton is using Sidekiq for plum, so that's worth taking a strong look at.

mcritchlow:
After a decent amount of research, it seems pretty clear that Sidekiq is being used by the [majority of the Rails community](http://blog.scoutapp.com/articles/2017/01/09/state-of-the-2017-rails-stack), and I've noticed new Hydra apps are using it s well. It is also noticeably faster and comes with an option for paid support (along with additional features) that we could evaluate in the future.

So perhaps we should adjust this ticket to adding in Sidekiq as the queue solution. @ucsdlib/developers  -- does that reasonable to you all?

lsitu:
@mcritchlow Sidekiq looks good and I am fine with it.

mcritchlow:
@lsitu since this overlaps with your work in #23 and #15 do you think it makes sense to break out any additional configuration/etc. into the scope of this ticket?

lsitu:
@mcritchlow Yes. It sounds good to break it out any additional configuration/etc. into the scope of this ticket. Thanks.

lsitu:
@mcritchlow I am able to setup Resque/Resque Pool and try to use it for CSV/Batch Ingest. I don't see any warnings/errors for ingest and it seems like that the records and thumbnails are showing up faster and in the orders of ingest, while using Sidekiq the order is messing up and often I need to wait longer time for all the records to be showing.
Would you like me to create two different PRs for Resque and Sidekiq configuration so that we all can evaluate them a little bit before making a decision?

mcritchlow:
[...]

PR #25, opened 2017-01-18
Fixes configuring Resque as a queue option. #17

lsitu:
https://github.com/ucsdlib/horton/issues/17
Start Resque:
`bundle install` 
`RUN_AT_EXIT_HOOKS=true TERM_CHILD=1 QUEUE=* rake environment resque:work`
Or
`RUN_AT_EXIT_HOOKS=true TERM_CHILD=1 rake resque:pool`


mcritchlow:
I'm wondering if we should drop this path. Another point of interest is there hasn't been an official release of Resque in [quite a while](https://github.com/resque/resque/releases)

Given the potential supported plan of Sidekiq and active development, any reason to not make the call to use Sidekiq?

lsitu:
@mcritchlow I think this PR is just for research and comparing. I would close the PR now since Sidekiq is under active development and more widely accepted by the community.

PR #31, opened 2017-02-02
Fixes #13 - Add Continuous Integration

### LINKED ISSUE #13: Add Continuous Integration and related plugins/services

VivianChu:


VivianChu:
@ucsdlib/developers - Please review, comment, etc... Thanks

circle.yml
```diff
@@ -0,0 +1,38 @@
+general:
+  artifacts:
+    - "tmp"
+
+machine:
+  node:
+    version: 6.1.0
+  ruby:
+    version:
+      2.2.3
```
mcritchlow: Can we change to 2.4?

circle.yml
```diff
@@ -0,0 +1,38 @@
...
+    - wget http://projects.iq.harvard.edu/files/fits/files/fits-1.0.5.zip -O fits-1.0.5.zip
+    - unzip fits-1.0.5.zip -d /tmp/
+    - sudo chmod 755 /tmp/fits-1.0.5/fits.sh
+    - sudo ln -s /tmp/fits-1.0.5/fits.sh /bin/fits.sh
+  override:
+    - bundle check --path=vendor/bundle || bundle install --path=vendor/bundle --jobs=4 --retry=3
+  post:
+    - bundle exec rake hydra:test_server:
+        background: true  
+    - sleep 10 && wget --waitretry=5 --retry-connrefused -v http://localhost:8986/rest
```
mcritchlow: It looks like this checks for Fedora to be started and assuming Solr succeeds. That's probably true the majority of the time, but in theory starting up Solr could fail as well. Also if for some reason CircleCI is particularly slow, I wonder if we might go past the 5 retries?

What if we created a little script in our horton bin directory that we can call like:
`bin/test_server_wait`

And then populate that script with checks for Fedora and Solr. Something like the little script below. I'm not sure what the wget equivalent of this is, I'm just a bit more familiar with curl.

@jhriv - may actually have a cleaner way of doing this we could try also

```
#!/usr/bin/env sh

printf 'Starting Solr.'
until $(curl --output /dev/null --silent --head --fail http://localhost:8985/solr); do
    printf '.'
    sleep 1
done

printf 'Starting Fedora.'
until $(curl --output /dev/null --silent --head --fail http://localhost:8986/rest); do
    printf '.'
    sleep 1
done
```

jhriv:
Besides a curl test, a netstat or ss call can be made to see if anything is listening on the port, but the service may not be ready as soon as the port is open.

ss -tl '( sport = :8985 )' | grep-q LISTEN
netstat -nl | grep -q 8985

VivianChu:
@mcritchlow - The script has been added and circleci config has been updated.  It's working fine.  However, bundle install failed with ruby version 2.4.

mcritchlow:
@VivianChu very strange about 2.4 not working, but we can always circle back to updating Ruby as a separate task.

mcritchlow:
@lsitu or @hweng any concerns?

lsitu:
👍

lsitu:
@mcritchlow I upgrade ruby to 2.4.0 and it works in my local environment.

VivianChu:
@mcritchlow - Here's the error message in here https://circleci.com/gh/ucsdlib/horton/66
make failed, exit code 2

Gem files will remain installed in
/home/ubuntu/horton/vendor/bundle/ruby/2.4.0/gems/json-1.8.3 for inspection.
Results logged to
/home/ubuntu/horton/vendor/bundle/ruby/2.4.0/extensions/x86_64-linux/2.4.0-static/json-1.8.3/gem_make.out

An error occurred while installing json (1.8.3), and Bundler cannot
continue.

mcritchlow:
@VivianChu @lsitu Interesting. It looks like there was/is an issue with that version of the JSON gem and Ruby/Rails compatibility https://github.com/rails/rails/issues/27450#issuecomment-269500870

lsitu:
@mcritchlow @VivianChu Yes, it's strange. Maybe we can try 2.3.1 for now or just circle it back later?

mcritchlow:
How about we get this merged so we can then get the other PR's merged, and then I'll setup another ticket we can work on next Sprint for getting Ruby updated?

lsitu:
@mcritchlow Yes. It sounds good and I'll merge it now.

PR #39, opened 2017-02-16
Fixes #34 - Adding hound and rubocop config and coveralls gem

### LINKED ISSUE #34: Add Coveralls/Hound CI reporting
Following the work @hweng did on https://github.com/ucsdlib/billing-ill/issues/98

Please add Coveralls and Hound CI to this repo.

For linters, let's start with rubocop. We can also discuss CSS/JS options, but it doesn't have to be in scope for this ticket.

hweng:


hweng:

.hound.yml
```diff
@@ -0,0 +1,10 @@
+scss:
+  enabled: false
+
+ruby:
+  config_file: .rubocop.yml
+
+jshint:
+  enabled: false
+
+fail_on_violations: true
```
mcritchlow, line 10: 👍 is everyone else good with failing on violations?

.rubocop.yml
```diff
@@ -0,0 +1,37 @@
...
+  Enabled: true
+  EnforcedStyle: single_quotes
+
+Style/Documentation:
+  Exclude:
+    - 'spec/**/*'
+    - 'app/**/*'
+
+Style/Next:
+  Enabled: false
```
mcritchlow: Why was this disabled?

.hound.yml
VivianChu: 👍

lsitu: 👍

.rubocop.yml
hweng: Use `next` to skip iteration instead of a condition? It is fine to me if you want to stick to that.

mcritchlow: I was just wondering why it was selectively disabled, since it is a default (and default syntax convention in ruby). I would vote for having it enabled, but it's up to you all of course. @lsitu  and @VivianChu any preference? To give context, we're talking about the following:

http://rubocop.readthedocs.io/en/latest/cops_style/#stylenext

hweng: @mcritchlow Yeah, here are some comments from other folks including bbatsov about this cop: https://github.com/bbatsov/rubocop/issues/1238

VivianChu: It's fine to have it enabled.  I got it enabled for dmr and use next to skip iteration.

lsitu: I am fine to have it enable too.

mcritchlow: @hweng yeah, there are some examples in that issue I could see being annoying to have described as a "failure", since some of those are subjectively better syntax than using next.

Generally it seems like good practice to use next as a guard clause, though if it got annoying or we had use cases like the ones mentioned crop up, we could always disable again or experiment with the MinBodyLength setting? Just a thought.

hweng: @mcritchlow Sounds fine to me, I will enable it now as you all voted it.
Just provide some of other folk's experience so that we could realize both sides of it.
And as you mentioned, we could always disable it if it is getting annoying.

hweng:
@ucsdlib/developers The style/next cop is enabled. Please review and merge it after the CircleCI tests passed. Thanks!

mcritchlow:
👍

mcritchlow (approved):
Approving per discussion with @ucsdlib/developers

PR #49, opened 2017-02-28
Fixes #47 - Update A/V streaming viewers for streaming from the derivatives filestore.

### LINKED ISSUE #47: Update A/V streaming viewers
### Descriptive summary

Decision was made in #11 to use the default Hyrax persistence strategy for derivatives. As a result, A/V streaming configuration/viewers and any related tests should be adjusted as necessary.

### Related work

https://github.com/ucsdlib/encrypted-stream-name/issues/5

lsitu:
Fixes #47 ; refs #47

Streaming video/audio from the default derivatives filestore instead of Fedora.

* Removed implementation for Fedora A/V derivative persistency
* Updated A/V streaming viewers to stream from the derivatives filestore 
* Added tests for the A/V viewers


mcritchlow:
👏 

A general question on file naming convention I see in this and the encrypted stream PR. Why are mp3 and mp4 filename suffixes added to their respective files?

For example I see: `filename-mp3.mp3` and `filename-mp4.mp4`

Why are these not: `filename.mp3` and `filename.mp4` ?

lsitu:
@mcritchlow Good question! I think a short answer is just the Hyrax naming convention that are ported from CC, https://github.com/projecthydra-labs/hyrax/blob/master/app/services/hyrax/derivative_path.rb#L55, which is the result of "#{destination_name}.#{destination_name}". The variable destination_name here is mp3, mp4 etc, which is not so straight forward by the the meaning of the word itself but the A/V derivative created does follow the pattern.

mcritchlow:
I see. I guess I could see something like `filename-audio.mp3` or `filename-video.mp4` similar to `filename-thumbnail.jpeg` that I've seen somewhere in the docs. I guess it's the repetitive use of the extension that feels strange. But thanks for clarifying for us.

mcritchlow:
👍

VivianChu:
👍

hweng:
👍

hweng (approved):

PR #65, opened 2017-03-16
Feature/identifier

### LINKED ISSUE #62: Implement Identifier support
### Descriptive summary

Following a meeting with @ucsdlib/domm there was agreement to model identifiers as individual predicates, as opposed to the custom URN syntax we had previously documented. 

@GregReser will be identifying and documenting the predicates to use as a starting point

@lsitu will implement these predicates and related support for ensuring the distinction between public/private identifiers are carried through from DAMS4.

### Related work

Link to related tickets or prior related work here. TBD

lsitu:
Fixes #62 ; refs #62
Implement Identifier support for objects and collections.

* Added namespace and properties proposed for identifiers to the data model. 
* Implement Identifier support for objects and collections.
* Removed files that are no long needed for collection presenter overriding.


app/forms/collection_edit_form.rb
```diff
@@ -8,6 +8,12 @@ class CollectionEditForm < Hyrax::Forms::CollectionForm
   delegate :brief_description, :general_note, :location_of_originals, :table_of_contents, to: :model
   delegate :resource_type, :language, to: :model
 
+  IdentifierSchema.properties.each do |prop|
```
mcritchlow: This code block is duplicated in the object form, and to a slightly lesser extent the solr document and show presenters. Would it make sense to extract the shared code?

app/forms/hyrax/object_resource_form.rb
```diff
@@ -28,6 +35,8 @@ def self.model_attributes(attrs)
       attrs.dup.each do |key, value|
         if value.is_a? Array
           attrs[key].map! { |v| to_hash(v) }
+        elsif value.nil? || value.to_s.strip.blank?
```
mcritchlow, line 21: Would a call to `#compact` prior to enumerating the attributes remove the need for the nil/blank clause?

app/forms/collection_edit_form.rb
```diff
@@ -21,6 +27,8 @@ def self.model_attributes(attrs)
     attrs.dup.each do |key, value|
       if value.is_a? Array
         attrs[key].map! { |v| to_hash(v) }
+      elsif value.nil? || value.to_s.strip.blank?
```
mcritchlow, line 17: Would a call to `#compact` prior to enumerating the attributes remove the need for the nil/blank clause?

app/views/hyrax/collections/_form_required_information.html.erb
```diff
@@ -20,6 +20,7 @@
 
     <%= render "records/edit_fields/language", f: f, key: 'language' %>
 
+  <%= f.input :doi %>
```
mcritchlow, line 23: Is DOI specifically listed here for a reason, as opposed to other identifiers?

mcritchlow (commented):
This looks great 👍  Just a couple questions from my end.

app/forms/collection_edit_form.rb
lsitu: Good point! I've move one step further to refactor the codes to use schema to manage the properties across the model, the presenter, and the solr document, which make it easy to add/delete/update properties in just the schema definition class, see https://github.com/ucsdlib/horton/pull/65/commits/9725032dcd988e45b72e3efafd945b0aba15a888.

app/forms/collection_edit_form.rb
lsitu: It looks like`#compact` will return a new hash while we need to operate on the the attrs hash in place,  So it may not be efficient to loop throughout the k/v in the hash for other operations.

app/forms/hyrax/object_resource_form.rb
lsitu: It may not work with `#compact`, which will return a new hash (See comment above).

app/forms/collection_edit_form.rb
mcritchlow: Makes sense, would `#compact!` do in that case? if not, please disregard.

app/views/hyrax/collections/_form_required_information.html.erb
lsitu: It's only for collections. The full form for the identifiers are large while we don't know what will be used except, so I only put doi in the form now, and I think we can just add line here for the identifier we need in the future. See Greg's [#issuecomment-287090601](https://github.com/ucsdlib/horton/issues/62#issuecomment-287090601).

mcritchlow: Ah, ok. Thanks for the pointer to the comment 👍

app/forms/collection_edit_form.rb
lsitu: @mcritchlow I just try to apply `#compact` in https://github.com/ucsdlib/horton/blob/feature/identifier/app/forms/collection_edit_form.rb#L7 `attrs = super(attrs).compact` but I got a weird undefined method `with_indifferent_access' for "0":String

mcritchlow: Oh, how interesting. I wonder if the nil check (or compact alternative that seems to be problematic) is actually needed since you're calling [`super`](https://github.com/projecthydra/hydra-editor/blob/4da9c0ea542f7fde512a306ec3cc90380691138b/app/forms/hydra_editor/form.rb#L58-L64) which appears to sanitize/delete empty attributes?

That said, you've spent the time on this code, and this isn't worth dwelling on too much. It was just a thought. Don't want to unnecessarily slow down this PR going in :)

mcritchlow:
@ucsdlib/developers any other feedback here?

mcritchlow (approved):

app/forms/collection_edit_form.rb
lsitu: @mcritchlow Yes, just for sanitize/delete empty attributes but it looks more like a upstream issue that should do the work, while I am seeing empty triples inserted which is not a big issue. How about I removing that conditional check for now?

mcritchlow: Whatever you think is best. I agree it seems like the upstream super call "should" clean those empty attributes out. But, it also looks like between the super call and the current nil check additional Language attributes are getting added. Perhaps that's the source?

In any event, I'm completely good leaving your check in for now, or whatever you think best, since there's certainly nothing wrong with making sure we don't insert empty data (I hope) :)

VivianChu:
👍

hweng:
:+1:

PR #71, opened 2017-03-24
Fixes #66 - Finalize ucsd:Agent implementation.

### LINKED ISSUE #66: Finalize ucsd:Agent implementation
### Descriptive summary

A decision was made with @ucsdlib/domm to implement a single [ucsd:Agent class](https://docs.google.com/spreadsheets/d/1OjTm1Kuzo-An-THdpUkjwa5puul-b4Lx3SK0lB6rvl8/edit#gid=616154196&range=E200) that subclasses edm:Agent

Please update our existing Agent implementation to reflect this new modeling for ucsd:Agent.

### Rationale

A decision was made for a single subclass of edm:Agent, in place of an alternative approach to use multiple (foaf) classes for different Agent types.

### Related work

TBD

lsitu:
Fixes #66 ; refs #66

Finalize ucsd:Agent implementation.

* Refactored edm:Agent implementation
* Implemented ucsd:Agent model 
* Changed label/alternate label from multi-value to single value
* Updated lookup strategic for ucsd:Agent by label and alternate label
* Applied presenter pattern for local authorities display
* Added additional tests for more test coverage


app/models/authority.rb
```diff
@@ -3,35 +3,48 @@ class Authority < ActiveFedora::Base
...
+  # xsd:anyURI
+  property :exact_match, predicate: ::RDF::Vocab::SKOS.exactMatch, multiple: true
 
   validates :label, presence: { message: 'Label is required.' }
   validates_with LabelExistsValidator
+  validates :exact_match, url: true, allow_blank: true
 
   def self.is_authority?(record)
-    record.is_a?(Authority) || record.class.ancestors.include?(Authority) || record.is_a?(Agent) || record.is_a?(Concept)
+    record.is_a?(Authority) || record.class.ancestors.include?(Authority) || record.is_a?(UcsdAgent) || record.is_a?(Concept)
```
mcritchlow: Does a UcsdAgent record not respond to the `record.class.ancestors.include?(Authority)` conditional? I guess I'm curious why we need the explicit checks for agent and concept.

app/models/ucsd_agent.rb
```diff
@@ -0,0 +1,30 @@
+# UcsdAgent is a subclass of EDM Agent (http://www.europeana.eu/schemas/edm/Agent)
+class UcsdAgent < Agent
+
+  configure type: ::UcsdTerms.UcsdAgent
+
+  property :agent_type, predicate: ::UcsdTerms.agent_type, multiple: false
+  property :has_orcid, predicate: ::RDF::URI.new("http://id.loc.gov/vocabulary/identifiers#orcid"), multiple: false
```
mcritchlow: Did @ucsdlib/domm decide the property would be called `has_orcid`? That seems strange to me, as it reads more like a boolean method name or conditional check. Could we just use `orcid`?

app/models/authority.rb
```diff
@@ -3,35 +3,48 @@ class Authority < ActiveFedora::Base
...
   end
 
   protected
+    def index_authorities(solr_doc, field_name, authorities)
+      authorities.each do |auth|
+        index_authority_field solr_doc, field_name, auth
+      end
+    end
+
+    def index_authority_field(solr_doc, field_name, field_value)
```
mcritchlow, line 35: I suspect we may end up wanting faceting implemented for some of these. But that can be a later discussion.

app/models/authority.rb
```diff
@@ -3,35 +3,48 @@ class Authority < ActiveFedora::Base
...
       obj = obj.id if obj.is_a? ActiveTriples::Resource
-      if obj.to_s.start_with?(ActiveFedora.fedora.host)
+      obj = begin
         ActiveFedora::Base.find(obj.split("/")[-1]).label
-      else
-        obj.to_s
-      end
-  end
+      rescue
+      end if obj.to_s.start_with?(ActiveFedora.fedora.host) || obj.to_s.start_with?(Rails.configuration.authority_path)
```
mcritchlow: I think we need a newline/break between `end if`? Or was this intended to be an `else` condition? As is, I'm not sure how to interpret the logic.

app/models/authority.rb
```diff
@@ -3,35 +3,48 @@ class Authority < ActiveFedora::Base
...
+      obj
```
mcritchlow: I'm not sure the explicit return is needed for `obj` though that might depend on clarification above. Potentially ignore :)

spec/controllers/dams_authorities_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
+describe DamsAuthoritiesController do
```
mcritchlow: Are these controller tests giving us additional test coverage that the feature tests are not? This is admittedly influenced by the recent trend of [Rails 5 removing controller test support out of the box](https://medium.com/table-xi/whats-up-with-rails-controller-tests-f0ece1fdd9f0#.of8ckh2qd). But honestly, I'm just curious because if there's anything we can do to make our test suite faster, while still maintaining coverage, I'm all for it

mcritchlow (commented):
A few comments and questions. But this looks really great. Nice work @lsitu 👍

app/models/ucsd_agent.rb
remerjohnson: @mcritchlow We could totally just use `orcid`. I was coming from a "be more explicit and include a verb" angle but realize that can sometimes instead create confusion.

lsitu: I'll change it to use `orcid` then. Thanks @remerjohnson.

app/models/authority.rb
lsitu: Yes, I think `record.is_a?(Authority) || record.class.ancestors.include?` should work and I will remove the other two explicit checks.

app/models/authority.rb
lsitu: Yes, I'll change it to use a new line with the `return obj unless ...` condition then.

app/models/authority.rb
lsitu: I'll change the syntax when adding the new line for condition above to make it more readable. Thanks.

spec/controllers/dams_authorities_controller_spec.rb
lsitu: I am not sure how these controller tests are counted by cover all but I can remove the class to see how it goes.

app/models/authority.rb
lsitu: I agree. We may need to discuss how to use the facet for the authority records and diferent it from those for objects when moving forward.

lsitu:
@mcritchlow I've made all the changes to rename has_orcid to orcid, and refactored the codes a little bit except that authority facet issue, see [commit](https://github.com/ucsdlib/horton/pull/71/commits/45887f0d7974297c1ad30f02238d95be5ba32167). I think keeping the test for the dams_authorities_controller is not a big deal and it's just okay to remove it.

mcritchlow (approved):
👍

mcritchlow:
@hweng - anything to add, concerns, etc.?

hweng:
@mcritchlow Looks fine to me. :+1:

Issue #53, opened 2017-03-01
Port WYSYWIG support


mcritchlow:
### Descriptive summary

Please port/recreate the ability for users with curator access to edit (and possibly create) new pages that they are able to edit themselves.

Examples under the Help menu of Digital Collections app.

While this feature is currently implemented in damspas in the following manner, this need not be the exact same implementation in Horton. If there are other options that may be a better fit, please feel free to recommend one. Ultimately, it is the functionality/feature that needs to migrate.

* https://github.com/ucsdlib/damspas/commit/6c5b536a86df4f251d815b2094f80d1580ab9433
* https://github.com/ucsdlib/damspas/commit/e3f2ba2efe6b25396791f8c0628d8b4d686d046a 

### Rationale

Curators would like to continue being able to update content for static pages without relying on the deploy cycle for the application.

hweng:
@mcritchlow  I figured out the problem that RSpec/FactoryGirl leaves record in test database after each test case is completed, which is messing up other tests: The "config.use_transactional_fixtures" should turn to "true"
https://github.com/ucsdlib/horton/blob/develop/spec/spec_helper.rb#L47

It works fine now after update the above line.

PR #72, opened 2017-03-27
Fixing the issue that test records left in test database.

hweng:
Fixing the problem that RSpec leaves record in test database after each test case is completed, which is messing up other tests which use ActiveRecord.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -82,7 +82,7 @@ group :development, :test do
   gem 'rails_autolink', '~> 1.1.6'
   gem 'unicorn', '~> 5.1.0'
   gem 'solr_wrapper', '>= 0.3'
-  gem 'database_cleaner', '~> 1.3'
+  gem 'database_cleaner', '~> 1.5.3'
   gem 'rails-controller-testing', '~> 0'
 end
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -722,7 +722,7 @@ DEPENDENCIES
   capybara (~> 2.6.0)
   coffee-rails (~> 4.2.0)
   coveralls
-  database_cleaner (~> 1.3)
+  database_cleaner (~> 1.5.3)
   devise
   devise-guests (~> 0.5)
   factory_girl (~> 4.0)
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -66,6 +66,10 @@ RSpec.configure do |config|
     end
   end
 
+  config.after :each do
+    DatabaseCleaner.clean
+  end
+
   config.before(:all, type: :feature) do
     # Assets take a long time to compile. This causes two problems:
     # 1) the profile will show the first feature test taking much longer than it


lsitu:
@hweng I don't see any test failures. What tests failed do you see? 
I think we are using database cleaner to manage the test data as what Hyrax/Sufia do: https://github.com/projecthydra-labs/hyrax/blob/master/spec/spec_helper.rb#L158-L163, which is different from running every example within a transaction by turning config.use_transactional_fixtures = true.

hweng:
@lsitu To clarify, It is for object created through ActiveRecord.  If you don't turn config.use_transactional_fixtures = true, The test record will be left in the test database and mess up other tests.
You can try it your local by create the object and tests using ActiveRecord, and you will see the problem.

lsitu:
Is it the test cases issue or a database cleaner issue that fails to ensure a clean state for testing,  see https://github.com/DatabaseCleaner/database_cleaner?

hweng:
@lsitu It took me a while to figure the issue, and I'd be happy to save your time If you haven't tried it on your local yet - you can look into the circleCI test failures which create the test records through ActiveRecord:   https://circleci.com/gh/ucsdlib/horton/121, 
you would see that they got failed because the test records didn't get clear up and messed up others.

And after I update config.use_transactional_fixtures = true, all tests have passed on my local.

Another you can try is to update the damspas app to turn the config.use_transactional_fixtures from "true" to "false", and you will see all the tests which create the object through ActiveRecord get failed.

I think it will help you understand the issues if you try it.

mcritchlow:
Looking at our existing configuration, it seems that while we do a clean when [starting the `:suite`](https://github.com/ucsdlib/horton/blob/develop/spec/spec_helper.rb#L50-L52), and we ensure before `:each` spec we [set the appropriate strategy](https://github.com/ucsdlib/horton/blob/develop/spec/spec_helper.rb#L60-L67), we do not explicitly clean between individual tests as we do with the[ ActiveFedora cleaner](https://github.com/ucsdlib/horton/blob/develop/spec/spec_helper.rb#L54-L58).

Between that, and the relevant example config on the [gem readme](https://github.com/DatabaseCleaner/database_cleaner/blob/master/README.markdown#rspec-with-capybara-example), I wonder if that might be related? Specifically, I'm wondering if something like the following would suffice as a separate block would ensure expected behavior:

```
config.append_after(:each) do
    DatabaseCleaner.clean
end
```

lsitu:
@hweng I think it's good to try it with the database cleaner configuration above that Matt suggested. And I believe the following code could fix the errors you see as well:
```
  config.after do
    DatabaseCleaner.clean
  end
```

If not, I think the code validation error could be fixed with sequence in FactoryGirl like `sequence(:code){|n| "code_#{n}" }` as well since the code is expected it to be unique for your test:
```
 6) PagesController authenticated PUT update with valid params updates the requested page
     Failure/Error: @page = create(:page)
     
     ActiveRecord::RecordInvalid:
       Validation failed: Code has already been taken
```

hweng:
@mcritchlow You are right that our current config has problem which don't clean between individual tests for object created by ActiveRecord.

To fix this issue, it turned out that just make "config.use_transactional_fixtures = true" would work.
We could certainly try the approach you suggested as well.

@lsitu I have to correct you, The source problem exists in the config which cause the test record left in the test database which mess up the other tests, that's why you see the failures that "Code has already been taken" and "count expected: 0 got: 2"  and etc.
The one you suggested "adding code sequence in FactoryGirl" won't fix the problems.

This problem will affect all the test objects created through ActiveRecord if we will add in the future, not just the Pages tests that I'm working on.

lsitu (changes requested):
I think the issue is that the change in the PR is conflicting with the current use of the DatabaseCleaner. Here is the document regarding how to use the setting with config.use_transactional_fixtures and the ways suggested to manage the test data: https://relishapp.com/rspec/rspec-rails/docs/transactions.

While inclining to follow the community and take advantage of the [DatabaseCleaner](https://github.com/DatabaseCleaner/database_cleaner) gem, I hope adding the missing configuration that Matt suggested will work for you. If not, I think changing the way to test it and/or manage the test data ourselves like deleting the records created will do the trick too. 

If you want to use the setting "config.use_transactional_fixtures = true", please remove the DatabaseCleaner codes to avoid the conflicts: [spec_helper.rb#L30](https://github.com/ucsdlib/horton/blob/develop/spec/spec_helper.rb#L30) ,  [spec_helper.rb#L56-L58](https://github.com/ucsdlib/horton/blob/master/spec/spec_helper.rb#L56-L58), [spec_helper.rb#L60-L67](https://github.com/ucsdlib/horton/blob/develop/spec/spec_helper.rb#L60-L67) etc. Thanks.

mcritchlow:
I concur with @lsitu's review.

Also, the [alternative block solution](https://github.com/ucsdlib/horton/pull/72#issuecomment-289651109) that he proposed might be even safer/better than what I put in my comment since it would run under other conditions as well (after suite & context)

hweng:
@lsitu, @mcritchlow  I agreed to try the other option: adding missing config for database cleaner in my last comment.
What i disagreed with you @lsitu is your suggestion adding in the individual test such as `sequence(:code){|n| "code_#{n}" }`, which won't fix the problem because the test records would still get accumulated and it would never get the expected results such as expect 0 count after destroy test. I hope you understand it.

I don't have strong opinion to use which way. Either turn "config.use_transactional_fixtures = true" or adding missing config for database cleaner,  Once the problem of "test records left in test database" got resolved, all tests will pass.

hweng:
Try the option:  turn config.use_transactional_fixtures to be true:
https://github.com/ucsdlib/horton/commit/7a56d50cdf30396a65b65cdaefb3f2801f1b5780

Now all page tests passed:
https://circleci.com/gh/ucsdlib/horton/123

And I will try the other option: adding missing database cleaner config.

hweng:
Tried the other option: Adding missing database cleaner configuration:
https://github.com/ucsdlib/horton/commit/48c8a419d818c671490590b1321e77b1edcf798a

And all page tests got passed too:
https://circleci.com/gh/ucsdlib/horton/124

hweng:
@mcritchlow @lsitu , Tested both approach, It turned out that either way would work, see above test results.

At this point I don't have strong opinion to use which, and it's totally fine for me to use the way "Adding more database cleaner configuration".

And I prefer to the following syntax because it follows the conventions of other databaseCleaner config in the same file https://github.com/ucsdlib/horton/blob/develop/spec/spec_helper.rb#L60:

```
config.after :each do |example|
    DatabaseCleaner.clean
  end
```

But again I don't have strong opinion about it, so let me know What you prefer to?

hweng:
Some people compared the two:
http://stackoverflow.com/questions/13234454/test-speed-activerecord-use-transactional-fixtures-vs-databasecleaner-strategy
We only have page transaction use of ActiveRecord in our codebase so far.

mcritchlow:
Thanks for running the tests to confirm both strategies work.

That's a really interesting post you reference. It's a little old, but not that old. How about this, I'll create another ticket to look into this more. Doing so would involve work outside the scope of this ticket, which is to get the test suite passing with AR data. In it, we can look into the pros/cons of each system and do some actual benchmarking tests of our own to see if one solution is more performant than the other given our needs. I'm all for making our test suite as fast as possible 👍 

For now though, since we already have the gem, I'd propose we keep using it with the supplemented configuration so you can wrap up #53 

How does that sound?

hweng:
@mcritchlow Sounds fine to me. The latest version of database_cleaner: https://rubygems.org/gems/database_cleaner/versions/1.5.3
And In horton it is still 1.3.

For DatabaseCleaner.clean loop, do you like the syntax I posted in the previous comment? Or do you like the other syntax instead?

mcritchlow:
By syntax are you referring to the use of the `|example|` block variable? Or putting the `:each` in parenthesis? Or are you referring to using `:each` as opposed to passing in no scope variable?

For the block variable, I don't think it's needed because it's not being used within the block. In the cases we're defining it, it's being used in the block.

For the method argument in parens vs not, I don't have much of an opinion. I think it can go either way and sticking with the existing convention seems good.

For `:each` vs nothing, if `:each` works (it does since the tests passed for you) then I think we're good to go as you wrote it.

hweng:
I meant to be consistent with the current syntax,  which is same syntax as https://github.com/ucsdlib/horton/blob/develop/spec/spec_helper.rb#L60,
just replace "before" with "after". Is that fine to you?

Or do you want to replace all of them with other syntax?

hweng:
Here it is, let me know if it looks fine to you?
https://github.com/ucsdlib/horton/blob/feature/rspec_config_update/spec/spec_helper.rb#L69

And let me know if you want to upgrade the database_cleaner to the latest version? And if to upgrade, do you want it in the same pull request or separate pull request?

mcritchlow:
Looks good! If you don't mind updating the gem in this PR that would be great. 

I think in general we need to be better about keeping gems up to date, thanks for offering.

hweng:
I don't mind. I will test the latest version of gem with pages tests first, and then update it in this pull request.

hweng:
@ucsdlib/developers - please review.

mcritchlow:
@lsitu can you take a peek when you have a chance? I think we need an approval from you per your initial review requesting changes.

LGTM 👍

lsitu:
@mcritchlow 👍 Looks good! Thanks @hweng

lsitu (approved):

PR #74, opened 2017-03-28
Add the subject import and standard input steam Excel ingest templates

remerjohnson:
The purpose of this PR is to add the "subject import" and "standard input stream" Excel templates. The reason we want to move them to `horton` is because DOMM and RDCP should be able to share these templates with researchers or anyone inside (or outside) the library, which we cannot do currently because [dams-metadata](https://github.com/ucsdlib/dams-metadata) is private. With this being in `horton`, we could point people to the URL. 

* Adding a directory in the root of the repo, which hold these Excel templates
* Important note: currently these are the DAMS4 Excel templates


mcritchlow:
I really like the idea of this. Though I'm wondering why we'd put dams4 spreadsheets here instead of the damspas repo? I can see a scenario very soon where we'll have both DAMS4 and 5 spreadsheet templates. Does it make sense to have dams4 in damspas and dams5 in this repo?

remerjohnson:
Greg and I were just talking about the D4/D5 divide for templates. That would be a very obvious solution! :+1:

mcritchlow (approved):
👍

lsitu:
👍

PR #77, opened 2017-03-30
Amend PR Procedure

mcritchlow:
This follows a conversation with @hweng on clarifying the need to both:

a) push to the remote branch after rebasing and running tests locally
b) using the `--force-with-lease` option in the case where a remote branch already exists.

The issue with (b) may need a bit more investigation, since while this has normally worked fine in my experience, we ran into an issue just today with a feature branch of Huawei's that didn't work using the lease option. We'll test more and verify.

Also, when merging this PR, please use the "Squash and Merge" option and squash the `minor markdown syntax tweaks...` commit so that's not in the history.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -5,19 +5,20 @@ Please use the following process for creating a new [Pull Request](https://help.
 
 1. Create a feature branch for new work.
 2. Push commits to the feature branch, with appropriate test coverage (and tests passing).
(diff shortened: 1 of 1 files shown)


lsitu (approved):

lsitu:
👍

mcritchlow:
@VivianChu - now that you're back and this is yet to be merged, please take a look as well. Thanks!

VivianChu:
👍

PR #79, opened 2017-03-31
Fixes #53 - Adding WYSYWIG and page model, controller, views and migrate

### LINKED ISSUE #53: Port WYSYWIG support

hweng:
Fixes #53 ; refs #53

Adding WYSYWIG and page model, controller, views and migrate

* Added page model, controller, views and migrate
* Added rspec tests
* Added WYSYWIG support
* Added access control for pages action.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -11,9 +11,9 @@ class PagesController < ApplicationController
     @pages = Page.all
   end
 
-  # GET /pages/code/view
+  # GET /pages/slug/view
   def view
-    @page = Page.find_by code: params[:id]
+    @page = Page.find_by slug: params[:id]
   end
 
   # ---------------
@@ -30,7 +30,7 @@ class PagesController < ApplicationController
     @page = Page.new(page_params)
 
     if @page.save
-      redirect_to pages_url, notice: 'Page was successfully created.'
+      redirect_to pages_url, notice: t('pages.create_notice')
     else
       render action: 'new'
     end
@@ -47,7 +47,7 @@ class PagesController < ApplicationController
   # PATCH/PUT /pages/1
   def update
     if @page.update(page_params)
-      redirect_to edit_page_path(@page), notice: 'Page successfully updated.'
+      redirect_to edit_page_path(@page), notice: t('pages.update_notice')
     else
       render action: 'edit'
     end
@@ -56,7 +56,7 @@ class PagesController < ApplicationController
   # DELETE /pages/1
   def destroy
     @page.destroy
-    redirect_to pages_url, notice: 'Page was successfully destroyed.'
+    redirect_to pages_url, notice: t('pages.destory_notice')
   end
 
   private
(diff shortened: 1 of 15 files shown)


app/controllers/pages_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+  # ------------------------
+
+  # GET /pages
+  def index
+    @pages = Page.all
+  end
+
+  # GET /pages/code/view
+  def view
+    @page = Page.find_by code: params[:id]
```
mcritchlow: I think we should consider renaming `code` to `slug` to follow the [Semantic URL Slug definition](https://en.wikipedia.org/wiki/Semantic_URL#Slug) 

I know it's called `code` in the existing system, but this seems like a good opportunity to give that property a more accurate name?

app/views/pages/new.html.erb
```diff
@@ -0,0 +1,33 @@
...
+  </div>
+  <div class="field">
+    <%= f.label :title %><br>
+    <%= f.text_field :title %>
+  </div>
+  <div class="field">
+    <%= f.label :body %>
+    <%= f.text_area :body, size: '70x30' %>
+  </div>
+  <script src="//cdn.ckeditor.com/4.4.4/standard/ckeditor.js"></script>
```
mcritchlow: Is it possible to use the [latest and greatest version](http://ckeditor.com/download) of CKEditor?

app/views/pages/new.html.erb
```diff
@@ -0,0 +1,33 @@
...
+  <script>CKEDITOR.replace('page_body')</script>
+
+  <br/>
+  <%= f.submit 'Create Page', class: "btn btn-primary" %>
```
mcritchlow: Could we extract things like labels into our [internationalization file](https://github.com/ucsdlib/horton/blob/develop/config/locales/en.yml) for the application? It would be nice to follow that pattern for everything, since we're increasingly going to take advantage of the work in Blacklight and Hyrax in this context.

mcritchlow (changes requested):

hweng:
@mcritchlow All requested changes has been added to https://github.com/ucsdlib/horton/pull/79/commits/06770812616a7b0598a012e508b9eee4619c20ea,
Could you take a look and let me know if it looks fine to you. Thanks!

mcritchlow (approved):

VivianChu:
👍

lsitu:
👍

Issue #46, opened 2017-02-24
Create initial delimited import template


mcritchlow:
### Descriptive summary

Create an initial csv import template to be used for initial tool R&D.  

Focusing on:
1. hierarchical components
2. linked records
3. CSV vs TSV vs Excel - is there a preferred format that will make life easier for DOMM and Dev?

Deliver:
* An initial template
* An initial set of unit and acceptance tests

### Dependency
#111 

### Related work

Ticket migrated from https://github.com/ucsdlib/dams5-cc-pilot/issues/39

ghost:
@mcritchlow for "2. linked records", do you mean including URIs for concepts, like names and subjects?

mcritchlow:
@GregReser yeah, exactly.

ghost:
Would URI's be required?  We would need a pre-ingest tool to generate ARKs for all concepts already in the DAMS, then we would find URI's for the rest?

ghost:
Maybe the fastest way to completing this initial import tool would be to duplicate the functions of the current tool, with a few improvements.  For URIs, we could do something similar to our pre-OLR ingest subject import, but expand it to agents.

mcritchlow:
So that's a good question, and I think maybe gets at how the batch import tool fits into the overall DOMM workflow where you're presumably using something like Open Refine first. I'm not sure we need a separate tool. I think, similar to what we do now (though @lsitu would know better), we could do something like:

```
if cellValue is a URI
  try to lookup/match to an existing URI in the system
  if it exists, link to the existing URI
  if it doesn't exist. create a new local URI
if cellValue is a String
  try to find a URI using CV list (LCSH, Getty, etc.)
  if one is found, use it
  if not, create a new URI
```

Another variant on the String path is to just directly create a URI instead of doing remote lookup. This would be faster, of course. It would largely depend on whether you all would have done this ahead of time in OpenRefine or a similar tool. If you had, having our ingest tool do the lookup would be redundant and unnecessarily slow things down.

mcritchlow:
@GregReser - I think that's a good practical approach.

Just noting here that if we do choose to use Excel as the file format, instead of CSV/TSV, I checked with Alex and for a different tool he used the following gems were useful:

* https://github.com/weshatheleopard/rubyXL (we could just use this if we can stick to `.xslx`
* https://github.com/zdavatz/spreadsheet (we would need this if we have to support `.xls`

lsitu:
Thanks @mcritchlow for the information. 
[...]

PR #82, opened 2017-04-07
Fixes #81 - Batch import support for Excel format with object import template validation.

### LINKED ISSUE #81: Batch Import Support Update
### Descriptive summary

This ticket primarily addressed a number of issues outlined by @GregReser in his [comment](https://github.com/ucsdlib/horton/issues/46#issuecomment-291241448) via #46 

Summarized again here:

- Initial support for complex objects
- resolve relatedResource predicate(s) and Class #58
- resolve rights properties
        do we need copyright status or can it be inferred from rights URI?
- resolve Matt's question about identifier:roger record (should it be more generic, e.g., "identifier:catalog_record"
- add remaining agent predicates (it's a long list and I was waiting because it will make the template harder to read).

@lsitu In the object_import_template, I've been highlighting changes in yellow to make sure they you see them. I suggest you remove the yellow when you have implemented the change. This way we will all know what remains to be updated.

### Related work

#46

lsitu:
Fixes #81 ; refs #46, #81

Batch import support for Excel format with object import template validation.

* Implemented Batch Import support for Excel format.
* Added support for object import template validation.
* Refactored codes with the existing CSV Import functionality for more generic tabular data source support.
* Updated import template, predicates and properties.
* Implemented the support for the latest Excel import template.


app/views/batch_imports/new.html.erb
```diff
@@ -1,6 +1,6 @@
 <% provide :page_title, curation_concern_page_title(curation_concern) %>
 <% provide :page_header do %>
-  <h1>CSV Import</h1>
+  <h1>Batch Imports</h1>
```
mcritchlow: I'm seeing the title of this service in both plural and singular form. Unless this is necessary, we should probably settle on one? I'd vote for `Batch Import` `batch_import` etc.

lib/import/tabular_parser.rb
```diff
@@ -0,0 +1,78 @@
...
+            attrs.dup.each do |k, v|
+              attrs.delete k
+              splict_values(k, v, attrs)
+            end
+          end
+        end
+      end
+
+      # attributes in a record
+      def splict_values(key, values, attrs)
```
mcritchlow: I think I recall this naming typo from before?

lib/import/tabular_basic.rb
```diff
@@ -0,0 +1,37 @@
+module Import
+  module TabularBasic
+    private 
+      # valid header fields
+      def parse_headers(row)
+        headers = []
+        $i = 0
```
mcritchlow: Ruby uses the `$` prefix to denote global variables, which this is not. Is there an enumerable method that could be used here instead?

lib/import/tabular_basic.rb
```diff
@@ -0,0 +1,37 @@
...
+      def extract_field(header, cell, attrs = {})
+        val = field_value cell
+        return unless !val.blank?
+        attrs[header] ||= []
+        attrs[header] << val
+      end
+
+      # extract field value
+      def field_value(cell)
+        cell.nil? ? cell : cell.to_s.strip
```
mcritchlow: Do we want to have nil values? or would you prefer to have them always be strings? See related comment in `#extract_field`

lib/import/tabular_basic.rb
```diff
@@ -0,0 +1,37 @@
...
+          headers.each_with_index do |header, index|
+            extract_field(header, row[index], attrs)
+          end
+        end
+      end
+
+      # attributes in a record
+      def extract_field(header, cell, attrs = {})
+        val = field_value cell
+        return unless !val.blank?
```
mcritchlow: depending on your thoughts to the nil question in `#field_value` this could instead be something like: `return if val.empty?` ?

lib/import/tabular_importer.rb
```diff
@@ -0,0 +1,332 @@
...
+        # FIXME: file use properties
+        file_1_use = attrs.delete :file_1_use
+        file_2_use = attrs.delete :file_2_use
+
+
+        # FIXME: license, it's changed in upstream in the master branch with Hyrax 1.0
+        license = attrs.delete :license
+
+
+        # file path: when appearing ignore uploaded files
```
mcritchlow: Should this be its own method?

mcritchlow (commented):
This is a big PR, so I'm hoping @hweng and/or @VivianChu can take a look as well, as I may have missed something.

Overall though, this is a great step towards getting this significant feature in place 👍

app/views/batch_imports/new.html.erb
lsitu: Yes. Let's use `Batch Import` `batch_import` then.

lib/import/tabular_parser.rb
lsitu: Yep, sorry it's a typo again.

lib/import/tabular_basic.rb
lsitu: It's looks like enumerable method are more specific to a specific format like `row.cells.each` for Excel while CSV don't have that support. But row[i] supported by both formats to access a cell at location i.
How about removing the $ sign? I'll also update it use [].tap to replace the `headers` variable declaration.

lib/import/tabular_basic.rb
lsitu: Let's use `return if val.empty?` then.

lib/import/tabular_basic.rb
lsitu: It think it's fine with either way but I'll change it to empty string instead of nil so that it's consistent with the `return if val.empty?` syntax above.

lib/import/tabular_importer.rb
lsitu: Yes, I'll extract it out.

app/views/batch_imports/new.html.erb
lsitu: @mcritchlow It looks like controller `batch_import` is use by other purpose:
 `             email_batch_import GET|POST /batch_import/:id/email(.:format)                              batch_import#email
                  sms_batch_import GET|POST /batch_import/:id/sms(.:format)                                batch_import#sms
             citation_batch_import GET      /batch_import/:id/citation(.:format)                           batch_import#citation
          email_batch_import_index GET|POST /batch_import/email(.:format)                                  batch_import#email
            sms_batch_import_index GET|POST /batch_import/sms(.:format)                                    batch_import#sms
       citation_batch_import_index GET      /batch_import/citation(.:format)                               batch_import#citation
                batch_import_index POST     /batch_import(.:format)                                        batch_import#create
                  new_batch_import GET      /batch_import/new(.:format)
`
Maybe we need to use other still use terms or `batch_imports` for the controller, while `Batch Import` as text display in the UI?

lsitu: @mcritchlow I've got the single `batch_import` route path working, which is a little trivial when I tried to configure for it. So just disregard my message earlier which is the result of my local configuration.

lsitu:
@mcritchlow I've added the [commit](https://github.com/ucsdlib/horton/pull/82/commits/984e36148839c97eda5804861f830b75c7f9c84e) above to rename it to `Batch Import` and address your suggestions above. It's ready for review now. Thanks.

lib/import/tabular_basic.rb
lsitu: @mcritchlow There is no enumerable method `row.each` for Excel. It only define the`[](ind)` method that references the the value at cells[ind].

lsitu:
@ucsdlib/developers Please review, comments etc ...

app/views/batch_imports/_form_metadata.html.erb
```diff
@@ -9,11 +9,11 @@
         input_html: { class: 'form-control' } %>
 
     <div class="with-headroom" style="padding-bottom:25px;">
-        <label>CSV Metadata Source</label><br />
```
hweng: Could the things like labels and etc.  extracted into our internationalization file for the application? It would be nice to follow that pattern for everything, which follows the community and take advantage of the work in Blacklight and Hyrax in this context.

lsitu: It sounds good and it's done in [commit](https://github.com/ucsdlib/horton/pull/82/commits/f108d921a783f48d57256ca0e4a9ca1fa666428a). Thanks.

hweng:
:+1:

VivianChu:
👍

VivianChu (approved):

lsitu:
@ucsdlib/developers Thank you all for your reviews and comments!

PR #87, opened 2017-04-20
Fixed #64 - Upgraded to Hyrax 2.0.alpha from the github master branch.

### LINKED ISSUE #64: Upgrade to latest Hyrax
### Descriptive summary

Many commits have been introduced in Hyrax over the last few weeks.

Please:

- update our application to point to the latest.
- Confirm all tests still pass
- If any errors arise that seem to be upstream issues with Hyrax, please reach out and create issues in the Hyrax repository as needed.

### Rationale

In order to not fall significantly behind on the pace of change with Hyrax, we need to be vigilant about keeping up-to-date with the latest changes.

lsitu:
Fixes #64 ; refs #64

Migrated to the latest Hyrax 2.0.alpha from the github master branch. 

* Upgrade Hyrax.
* Fixed failed tests with code refactoring.


lsitu:
@ucsdlib/developers Please review, comments. The change are big and please make sure it works in your local environment with`bundle install` and `rake db:migrate`. Thanks.

app/mailers/application_mailer.rb
```diff
@@ -0,0 +1,4 @@
+class ApplicationMailer < ActionMailer::Base
+  default from: 'from@example.com'
```
mcritchlow: do we have configuration for this somewhere else? And does Hyrax make active use of ActionMailer for notifications? Essentially, I'm wondering about having a from address hard-coded here

app/models/ability.rb
```diff
@@ -17,18 +17,29 @@ def admin_permissions
...
+    can [:create, :edit], BatchImportItem
   end
 
   def curator_permissions
-    can [:read], curation_concerns
-    can [:read, :download], FileSet
-    can [:read], Collection
+    can [:read], :all
+    can [:download], FileSet
+    cannot :manage, :admin_dashboard
```
mcritchlow: why do we not want to allow curators to manage the admin dashboard? Are there features presented we don't want to provide access to?

mcritchlow (commented):
Just a couple questions that I don't see as blockers at all. This is a huge amount of work, nice job @lsitu 👍

app/mailers/application_mailer.rb
lsitu: @mcritchlow Yes, it looks like we can configure the [mailboxer.rb#17](https://github.com/ucsdlib/horton/blob/develop/config/initializers/mailboxer.rb#L7) for it. Would you like to use the dummy one (dams-support@ucsd.edu) from dams4 or just leave as-is for now?

app/models/ability.rb
lsitu: There are functions like admin set and workflows roles management bundled together in the admin-dashboard and I am not sure we want to expose them to curators or not. But I think we can enable the admin dashboard for curator now and re-examine it later. What do you think?

app/mailers/application_mailer.rb
mcritchlow: I'm curious what it's used for in Hyrax. In general, yes I'd say that dams-support address is probably fine, but it would be nice if we knew a bit more about the context.

app/models/ability.rb
mcritchlow: Yeah, I can see that. Good point.. My concern, on the other side, is if we're then disabling them from accessing features that we do want them to have access to. Maybe we should enable it for now, while we're learning, and decide on what to lock down once we have a better understanding of the features?

app/mailers/application_mailer.rb
lsitu: I think Hyrax use the mailboxer gem but adding the base class as an option to allow us to extend it for other mailing services that we may need. We can just remove it now and add it in later when we need other/different mailing services. Do you like to remove it for now?

app/models/ability.rb
lsitu: Yes. It sounds good and I'll remove that line to open the admin dashboard for the curator role for now.

lsitu:
@mcritchlow I've enabled admin dashboard for curators access and removed the ApplicationMailer template generated from the Hyrax app with its related files. We can add them in later when we found we need them. See commit https://github.com/ucsdlib/horton/pull/87/commits/ae7a7af21d8a3930d8d4b590221cb5bc996e805a.

mcritchlow:
@hweng and @VivianChu - this is a really big change. As @lsitu said, if you all can take a look and even try locally, it would be great. This is also a dependency on some other, new work, happening, so please look when you can. Thanks.

VivianChu:
@mcritchlow @lsitu - I tried it locally and it worked fine.  The tests also passed.  👍

hweng (approved):

PR #88, opened 2017-04-21
Fixes #55 - Set up sitemap and customized configuration

### LINKED ISSUE #55: Port Sitemap feature
### Descriptive summary

We need to recreate the sitemap feature we have in damspas, currently using the following gem [1].

If there is any built-in feature in Hyrax, perhaps we can use that and configure as needed.

If there is not, please port the sitemap feature to Horton.

1. https://github.com/ucsdlib/damspas/blob/develop/Gemfile#L16

A dependency on completing this work is a decision about Capistrano vs Ansistrano and getting the core of that setup. This will be in a separate ticket.

### Rationale

The sitemap feature is needed to support and enhance our SEO.

### Related work

Depends on #56

hweng:
Fixes #55

### DIFF
--- a/config/sitemap.rb
+++ b/config/sitemap.rb
@@ -1,5 +1,5 @@
 # Set the host name for URL creation
-SitemapGenerator::Sitemap.default_host = "http://library.ucsd.edu/hyrax"
+SitemapGenerator::Sitemap.default_host = "http://library.ucsd.edu/dc"
 SitemapGenerator::Sitemap.compress = :all_but_first
 
 SitemapGenerator::Sitemap.create do
@@ -22,14 +22,14 @@ SitemapGenerator::Sitemap.create do
       total = 0
       more_records = true
       solr_url = ActiveFedora.solr_config[:url]
-      puts "solr: #{solr_url}"
+      Rails.logger.info "solr: #{solr_url}"
       solr = RSolr.connect( :url => solr_url )
       while ( more_records )
         solr_response = solr.get 'select', :params => {:q => "has_model_ssim:#{record_type} AND read_access_group_ssim:public", :rows => rows, :wt => :ruby, :start => done, :sort => 'id asc'}
         response = solr_response['response']
         if done == 0
           if SitemapGenerator::Sitemap.verbose
-            puts "#{record_type}: #{response['numFound']} records"
+            Rails.logger.info "#{record_type}: #{response['numFound']} records"
           end
           total = response["numFound"]
         end
@@ -41,7 +41,7 @@ SitemapGenerator::Sitemap.create do
           id = rec['id']
           lastmod = rec['timestamp']
           if SitemapGenerator::Sitemap.verbose
-            puts "#{record_type}: #{id}, lastmod: #{lastmod}"
+            Rails.logger.info "#{record_type}: #{id}, lastmod: #{lastmod}"
           end
           add "#{record_path}/#{id}", priority: 0.9, :changefreq => 'monthly', :lastmod => lastmod
         end
@@ -52,7 +52,7 @@ SitemapGenerator::Sitemap.create do
         end
       end
     rescue Exception => e
-      puts e.backtrace
+      Rails.logger.info e.backtrace
     end
   end
 end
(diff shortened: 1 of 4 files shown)


config/sitemap.rb
```diff
@@ -0,0 +1,58 @@
...
+    :Collection => "collections"
+  }
+  resources.each do |record_type,record_path|
+    begin
+      rows = 100
+      done = 0
+      total = 0
+      more_records = true
+      solr_url = ActiveFedora.solr_config[:url]
+      puts "solr: #{solr_url}"
```
mcritchlow: Could we use the logger here instead of puts?

config/sitemap.rb
```diff
@@ -0,0 +1,58 @@
...
+      solr = RSolr.connect( :url => solr_url )
+      while ( more_records )
+        solr_response = solr.get 'select', :params => {:q => "has_model_ssim:#{record_type} AND read_access_group_ssim:public", :rows => rows, :wt => :ruby, :start => done, :sort => 'id asc'}
+        response = solr_response['response']
+        if done == 0
+          if SitemapGenerator::Sitemap.verbose
+            puts "#{record_type}: #{response['numFound']} records"
```
mcritchlow: Could we use the logger here instead of puts?

config/sitemap.rb
```diff
@@ -0,0 +1,58 @@
...
+        end
+        done += rows
+         
+        # output each record
+        @records = response['docs']
+        @records.each do |rec|
+          id = rec['id']
+          lastmod = rec['timestamp']
+          if SitemapGenerator::Sitemap.verbose
+            puts "#{record_type}: #{id}, lastmod: #{lastmod}"
```
mcritchlow: Could we use the logger here instead of puts?

config/sitemap.rb
```diff
@@ -0,0 +1,58 @@
...
+          add "#{record_path}/#{id}", priority: 0.9, :changefreq => 'monthly', :lastmod => lastmod
+        end
+    
+        # stop looping if this is the last batch
+        if done >= total
+          more_records = false
+        end
+      end
+    rescue Exception => e
+      puts e.backtrace
```
mcritchlow: Could we use the logger here instead of puts?

public/robots.txt
```diff
@@ -3,3 +3,5 @@
 # To ban all spiders from the entire site uncomment the next two lines:
 # User-agent: *
 # Disallow: /
+
+Sitemap: http://library.ucsd.edu/hyrax/sitemap.xml
```
mcritchlow: Since as you rightly pointed out, this won't be used until we hit production, and there's no real "configuration" here to speak of, perhaps we should just put the ultimate production URL in place now? Unless there's a good reason not to, as far as testing, etc. goes. 

`Sitemap: http://library.ucsd.edu/dc/sitemap.xml`

mcritchlow (commented):
LGTM 👍 

Just a couple comments about logging and, potentially, updating the robots.txt.

hweng:
@mcritchlow It is updated to use logger and the url is updated. Thanks for review!

mcritchlow:
A couple other things:
* We'll probably need to port the takedown static page as well: https://github.com/ucsdlib/damspas/blob/develop/config/sitemap.rb#L10 So that can go on the sitemap list as well.
* I wonder if our current sitemap is actually broken for the static pages. Take the takedown page above as an example. Our sitemap lists it as `/takedown`, but that actually will 404 since we have the /p/ prefix in the route, right? Shouldn't all the static pages be `/p/<page-title>`? So `/p/takedown`, `/p/contact`, etc.?

hweng:
@mcritchlow The url http://localhost:3000/about is working for me, I think  /about, /help and /contact are default in hyrax.
And yes, We have our own static pages creation mechanism. But We don't know yet what page title will be created in production, so I left it out now.

hweng:
@mcritchlow , See the updates - The dynamically generated page url is added and the default static page url is removed.

And also I double checked the solr query here is correctly, since the damspas solr query doesn't work for hyrax based project as the it is configured differently.

mcritchlow:
@hweng That's even better, nice! LGTM 👍

lsitu (approved):

lsitu:
👍

PR #91, opened 2017-04-28
Fixes #83 Implemented Breadcrumbs for Collection/Object hierarch

### LINKED ISSUE #83: Implement Breadcrumbs for Collection/Object hierarchy
### Descriptive summary

 As noted in #70 there is already breadcrumb support built in for Hyrax, however it is targeted more towards "The things that I have ingested" as opposed to the actual hierarchy of the collections and objects themselves.

### Rationale

This is a feature that is desired by the DLDP is to have objects include their parent collection in a breadcrumb interface. If possible, we should implement it to satisfy that feature requirement.

### Expected behavior

When viewing an object that is part of a Collection (say "Collection A") then breadcrumbs should appear on the object view page demonstrating this relationship.

### Actual behavior

The current behavior is nicely demonstrated via screenshots from @hweng in #70 

### Related work

#70

hweng:
Fixes #83

### DIFF
--- a/app/controllers/hyrax/object_resources_controller.rb
+++ b/app/controllers/hyrax/object_resources_controller.rb
@@ -50,5 +50,27 @@ module Hyrax
     def show_presenter
       ::ObjectShowPresenter
     end
+
+    def add_breadcrumb_relation
+      if action_name == 'edit' || action_name == 'show'
+        @document = ::SolrDocument.find(params[:id])
+        if !@document[:member_of_collections_ssim].blank?
+          collection_name = @document[:member_of_collections_ssim].first
+          collection_id = @document[:member_of_collection_ids_ssim].first
+          add_breadcrumb collection_name, hyrax.collection_path(collection_id)
+        end
+      end
+    end
+
+    def build_breadcrumbs
+      if action_name == 'show'
+        default_trail
+      else
+        add_breadcrumb t(:'hyrax.controls.home'), root_path
+        add_breadcrumb t(:'hyrax.dashboard.breadcrumbs.admin'), hyrax.dashboard_path
+      end
+        add_breadcrumb_relation
+        add_breadcrumb_for_action
+    end
   end
 end
--- a/spec/features/object_resource_spec.rb
+++ b/spec/features/object_resource_spec.rb
@@ -3,12 +3,15 @@
 require 'rails_helper'
 include Warden::Test::Helpers
 
-feature 'Create a ObjectResource' do
+feature 'ObjectResource' do
 
   let!(:admin_user) { FactoryGirl.create(:admin) }
   let!(:public_object_resource) { FactoryGirl.create(:object_resource, title: ["Public Object Title"], :user => admin_user) }
   let!(:campus_only_object_resource) { FactoryGirl.create(:campus_only_object_resource, title: ["Campus Only Object Title"], user: admin_user) }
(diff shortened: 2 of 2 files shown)


lsitu:
I test it and it looks good for Object view. Some question:
1. Click on the file set link on the bottom of the object viewer, I see inconsistent breadcrumbs:
screen shot - fileset

2. From the /catalog search result page, click on an object to the object viewer, I see no links/breadcrumbs back to the search result. Hyrax has a breadcrumb `Back to search results`.  
screen shot object

3. Click on a collection, I see breadcrumb `Your Collections`. Maybe it make more sense to have a breadcrumb to All `Collections` instead of `Your Collections` since the cortex is changed from `My Objects` now?
screen shot -collection

hweng:
@mcritchlow and @lsitu 

For #3, It is default collection page breadcrumb which is not in this ticket scope, so I didn't touch it. I think maybe it's better to ask Gabriela, what does she want to see of collection page breadcrumb?

For #2, The following screenshot is the default one from the /catalog search result page . Do you want to have modified breadcrumb:
`My Dashboard / Test Collection / Test Object`

Or to keep the default breadcrumb?
`Back to search results`

My understand of this ticket was to use the 1st choice, but let me know if you want to kept old one.

screen shot 2017-05-02 at 10 55 26 am

mcritchlow:
Great questions. Here's my take.

2. It is true that the scope of the ticket is for what Huawei has in this PR. That said, I can see @gamontoya and other users potentially wanting both since they serve different purposes. That is, the static hierarchical breadcrumb trail + a link back to search results (when available in context). This is actually something that might best be left to usability testing.. what do you think, @gamontoya ?

3.  This is an interesting point, and I suspect we are likely to want to remove ALL instances of "my" and "your" in the UI since we are not building an IR where individuals are uploading content. That said, as Huawei said it wasn't in scope for this ticket, and what's much more important is the end user/public view, for which this is not relevant.

My general take would be to leave the PR as-is and leave any further breadcrumb customization for usability testing, but I'd defer to @gamontoya there.

hweng:
Thanks @mcritchlow  For #1, File set page breadcrumb, it need to implement in file_sets_controller.rb. But it seems there is no such relationship metadata in file_sets that link back to parent id. It is one way but not both way, which in object there is link data to child files, but not vice verse:
screen shot 2017-05-03 at 9 31 40 am

screen shot 2017-05-03 at 9 31 05 am

mcritchlow:
@hweng - thanks for providing this screenshot and clarifying. I would opt for not bringing fileset breadcrumbs in scope for this ticket. This is an entirely new concept, and so I don't think it's technically in scope for what folks are expecting from breadcrumbs. I would vote to leave such a thing for usability testing and/or design.

lsitu:
@gamontoya I agree with @mcritchlow to leave the PR as-is. I just brought up the issues that are related to the breadcrumbs for your reference and I think we can create a new ticket if you want any changes. What do you think?

VivianChu:
👍

gamontoya:
@mcritchlow @hweng I agree with you all. Leave the PR as it is and we can ask for customization during or after an accessibility phase for DAMS5.

mcritchlow (approved):
Given feedback from @gamontoya and 👍 from @lsitu and @VivianChu , I'll add an approval.

Issue #92, opened 2017-05-01
Port Metadata-Only feature


mcritchlow:
### Descriptive summary

DAMS4 has a metadata only 'state' that is used for objects that we are allowed to make publicly available, but only the metadata, not any associated files.

If there is a newer/better means to accomplish this feature given the new Hyrax framework, let's discuss and implement that.

### Rationale

This is an existing DAMS access control feature that must be migrated.

### Expected behavior

Loading an object from the search results view, or via direct link, that has a metadataOnly value for rightsOverride (assuming agreement in #84 ), should only display the metadata for that object. No files should be visible or accessible from a public user.

Example in DAMS4: *needed for reference*

### Actual behavior

not implemented. At the moment the user would see the metadata and the files for the given object.

### Related work

#84 #85

lsitu:
@mcritchlow Hyrax has the visibility_badge that marks an object's state/visibility, which is triggered by the the visibility implemented in https://github.com/ucsdlib/horton/issues/84. And the file download is more or less related to access control too. So it looks like the feature here is a little bit overwrap with ticket https://github.com/ucsdlib/horton/issues/84. But I think we can evaluate what else we still need to port once ticket https://github.com/ucsdlib/horton/issues/84 is done.

Issue #85, opened 2017-04-14
Mockup new Visibility form


mcritchlow:
### Descriptive summary

@ucsdlib/domm could use a mockup/screenshot for what the visibility part of the form would look like when the new visibility values and descriptions are added in. This would be purely a web mockup, not an actual implementation, which is slated for #84 

The mockup should include the following:

1. Add new visibility options below the defaults: Suppress Discovery, Metadata Only, Culturally Sensitive
2. Add boilerplate helper text for each option. May need help from @ucsdlib/domm on this one.
3. ..?

### Rationale

It would be helpful to have a visual when discussing rights, access control, license, etc.. It will also help ensure we are only adding to the existing Rights/Access Control/Visibility design and behavior provided by Hydra.

### Related work

Documentation:
* https://lisn.ucsd.edu/display/DOMM/Revised+Rights+Metadata+Structure+Recommendation+2017
* https://docs.google.com/spreadsheets/d/134Hxqafmz70X0glD3T98-pLMUwOK18oftvdSWOYdJ2k/edit#gid=616154196

#84

lsitu:
@ucsdlib/domm What boilerplate helper text would you like for Suppress Discovery, Metadata Only, and Culturally Sensitive in item#2 under "Descriptive summary"?

lsitu:
@ucsdlib/domm I've moved ahead to mockup the visibility form and we can change the  boilerplate helper text later at any time. Here are the screenshots:

- Visibility component
<img width="396" alt="screen shot - visibility options" src="https://cloud.githubusercontent.com/assets/2430784/25191704/536bd4f6-24e5-11e7-9590-dfc34fcd542a.png">

- Batch Import Form
<img width="1437" alt="screen shot - batch import" src="https://cloud.githubusercontent.com/assets/2430784/25191832/b7a9b3b6-24e5-11e7-9f2d-4a643e7178c7.png">

- Object Edit Form
<img width="1440" alt="screen shot - edit form" src="https://cloud.githubusercontent.com/assets/2430784/25191846/c6fa23aa-24e5-11e7-9218-6534b062103f.png">

arwenhutt:
@lsitu 
This is helpful, thanks!
A rough pass at the description text:
- Suppress Discovery - exclude from searching and discovery but accessible via direct link
- Metadata Only - metadata record is available but files are suppressed
- Culturally Sensitive - files available after click through to indicate that materials may be of a sensitive nature

[...]

Issue #59, opened 2017-03-02
Port Suppress Discovery feature


mcritchlow:
### Descriptive summary

DAMS4 has a suppressDiscovery 'state' that is used for objects such as OpenTopography. We need this feature to exist in DAMS5/Horton as well.

If there is a newer/better means to accomplish this feature, let's discuss and implement that.

### Rationale

In addition to OpenTopo, we are very likely to have additional research data use cases for this soon, after speaking recently with @hjsyoo 

### Related Work

#85 #84

Issue #93, opened 2017-05-02
Upgrade to latest Hyrax


mcritchlow:
### Descriptive summary

Please update Hyrax to the latest version/commit in 2.0-alpha.

### Rationale

Staying up to date with Hyrax development.

hweng:
@mcritchlow After upgraded, I got 73 test failure. Basically they all point to `Failure/Error: @search_state ||= Hyrax::SearchState.new(self)`

Looking into hyrax and horton, I think that's because there were the exactly same methods copied over to our local app, and the #search_state method was removed from hyrax recently:
https://github.com/projecthydra-labs/hyrax/commit/e23b9c2e86ef177543d01a64ff7184d309ffc76c

For example the failure:
 ```
72) Create a UcsdAgent a logged in user should be able to create agent that links a closeMatch URL
      Failure/Error: @search_state ||= Hyrax::SearchState.new(self)

      ActionView::Template::Error:
        wrong number of arguments (given 1, expected 2..3)
      # ./app/controllers/hyrax/controller.rb:38:in `new'
      # ./app/controllers/hyrax/controller.rb:38:in `search_state'
      # ./app/views/catalog/_search_form.html.erb:2:in `block in _app_views_catalog__search_form_html_erb__589929521283573638_70152744767900'
      # ./app/views/catalog/_search_form.html.erb:1:in `_app_views_catalog__search_form_html_erb__589929521283573638_70152744767900'
      # ./app/controllers/application_controller.rb:13:in `anonymous_user'
      # ./spec/support/features/session_helpers.rb:7:in `sign_in'
      # ./spec/features/ucsd_agent_spec.rb:14:in `block (3 levels) in <top (required)>'
      # ------------------
      # --- Caused by: ---
      # ArgumentError:
      #   wrong number of arguments (given 1, expected 2..3)
      #   ./app/controllers/hyrax/controller.rb:38:in `new'
```

**Comment by hweng** [comment]
And also got 7 errors in batch_import_job.rb, such as 
```
7) BatchImportJob #perform Excel XL metadata source when visibility is passed sets public read access
     Failure/Error: status << work_actor.create(env)

     ActiveJob::SerializationError:
       Unsupported argument type: RDF::URI
     # ./app/jobs/ingest_work_job.rb:30:in `block in perform'
     # ./app/jobs/ingest_work_job.rb:15:in `each'
     # ./app/jobs/ingest_work_job.rb:15:in `perform'
[...]

PR #108, opened 2017-05-10
Fixed #84, #85, #92, #59 - Implemented the new rights metadata model with local customized visibilities.

### LINKED ISSUE #84: Implement new Rights metadata model
### Descriptive summary

While it was designed to simplify rights and access controls, the new rights management strategy and metadata model is significantly different then what we implemented in 4.

Need a full spec including:
- list of properties
- associated application behavior (verified with the library rights policy group)
- controlled value lists or validation rules for each property
- properties and validation rules for entry via the CSV ingest stream
- wireframe for entry via the MARC/AT ingest stream

Central to this, and a prerequisite before finalization of spec:

- analysis of existing Hyrax access controls (dev ticket?)
- depending on implementation the access controls related to QA and release to final access state (public, ucsd only, etc.)

### Related Work

#46

lsitu:
Fixes #84, #85, #92, #59

Implemented the new rights metadata model with support for the edit form, Batch Import and UI display.

* Implemented the new rights metadata model .
* Ported suppress-discovery feature.
* Ported the metadata-only feature.
* Added the thumb-restricted icon for culturally-sensitive items display.


lsitu:

app/models/file_set.rb
```diff
@@ -1,5 +1,28 @@
 # Generated by hyrax:models:install
 class FileSet < ActiveFedora::Base
   include ::Hyrax::FileSetBehavior
+
+  property :rights_override, predicate: ::RDF::URI.new("http://pcdm.org/2015/06/03/rights#rightsOverride"), multiple: false do |index|
+    index.as :stored_searchable
+  end
+
+  def visibility
```
mcritchlow: There is a bit of duplication between the definition of visibility in FileSets and Objects. Perhaps that's ok for now as this is the first implementation, and just something to keep an eye on for the future as a possible refactoring?

app/controllers/hyrax/downloads_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    # Loads the file specified by the HTTP parameter `:file`.
+    # If this object does not have a file by that name, return the default file
+    # as returned by {#default_file}
+    # @return [ActiveFedora::File, String, NilClass] Returns the file from the repository or a path to a file on the local file system, if it exists.
+    def load_file
+      file_reference = params[:file]
+      return default_file unless file_reference
+
+      icon = icon_path(params[asset_param_key], true) if file_reference == 'thumbnail'
+      return icon if !icon.nil?
```
mcritchlow: I'm not sure this needs the nil check. I think you may be able to just say

`return icon if icon`

You might also be able to use a return pattern such as the following for the rest of this method (just a thought):

```
result =
  if condition
    x
  else
    y
  end
```

app/controllers/hyrax/downloads_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    # use static thumbnails for visibility metadata only and culturally sensitive objects
+    # @param [string] asset_id
+    def icon_path(asset_id, absolute_path=false)
+      file_set = ActiveFedora::Base.where("id:#{asset_id}").first
+      return if (can? :edit, file_set) || file_set.rights_override.blank?
+      # retrieve icon
+      return ::ThumbnailPathService.icon_file_path(VisibilityService.visibility_value(file_set.rights_override)) if absolute_path
+      ::ThumbnailPathService.icon_path(VisibilityService.visibility_value(file_set.rights_override))
```
mcritchlow: What is the scenario where this last statement gets called? I guess I'm wondering if the path will ever not be absolute for our special icons, and whether this fallback will ever get called.

app/models/solr_document.rb
```diff
@@ -28,6 +28,18 @@ class SolrDocument
...
+  def visibility
+    if rights_override?
+      @visibility ||= VisibilityService.visibility_value(self[Solrizer.solr_name('rights_override')].first)
+    else
+      super
+    end
+  end
+
+  def rights_override?
+    !self[Solrizer.solr_name('rights_override')].nil? && !self[Solrizer.solr_name('rights_override')].first.blank?
```
mcritchlow: I think we can start making use of some newer syntax available in Hyrax for loading solr properties. In this case the `fetch` option might be nice. I tried to [document this](https://mcritchlow.github.io/hydra/development/2017/05/09/advanced-hydra-camp-day-1.html) from the Hydra Camp (see the Pairing Musical Work section)

mcritchlow (commented):
This looks really good, a really big set of features to land in one PR. Thanks @lsitu! Just a couple comments, but LGTM 👍

hweng:
Looks :+1: to me, and I agree with @mcritchlow comments above.

app/models/solr_document.rb
VivianChu: @mcritchlow - The newer syntax is really nice.  Overall this PR looks 👍 .  Would you like me to merge it or wait for Longshou to update it using fetch option?

mcritchlow: @VivianChu - I think I'd like to give @lsitu time to review and offer any thoughts on his end. Otherwise though, like I said it looks good to me.

app/models/file_set.rb
lsitu: Agree. The method `def visibility` is a duplication but yes, we can keep an eye on it for a possible future refactoring.

app/controllers/hyrax/downloads_controller.rb
lsitu: I think this will get called from the FileSet viewer for public access of the thumbnails in the case of culturally sensitive and metadata-only visibilities. The DownloadController need to access the local thumbnail/derivative filestore directly.

app/models/solr_document.rb
lsitu: I'll refactor it to use the new pattern then.

app/controllers/hyrax/downloads_controller.rb
lsitu: Yep, it should work with `return icon if icon. Let's go for it.

lsitu:
@ucsdlib/developers I've updated the PR to use the new pattern to fetch field values from solr_doc as @mcritchlow suggested. Please review, comments ... Thanks.
 
BTW, It seems like that the develop branch of the current horton codebase is broken with error "uninitialized constant ActionView::CompiledTemplates::Flipflip", an obvious`Flipflop` typo. And I can't rebase it. I think we must have been upgraded to a wrong hyrax version recently! Does anyone see that error? 

Here are the steps to replicate it after checking out the latest develop branch from https://github.com/ucsdlib/horton:
From the [dashboard](http://localhost:3000/dashboard?locale=en), click on menu`Works` on the middle of the left panel (or http://localhost:3000/dashboard/my/works?locale=en).

hweng:
@lsitu The version didn't change which is still '2.0.0.alpha'. It was upgraded to the latest commit at that point: projecthydra-labs/hyrax@562b8ff,  see https://github.com/ucsdlib/horton/pull/109,
as the ticket required:  https://github.com/ucsdlib/horton/issues/93,  
And all our local tests passed.

As you know the hyrax is keeping on developing, it is not surprised if it is conflict with our local development and we need to stay up to date with Hyrax development. You may upgrade to the latest commit from hyrax master branch to see if it works fine to you.

lsitu:
@hweng It was fixed 14 days ago, just shortly after the hyrax commit that are used for horton upgrade: https://github.com/projecthydra-labs/hyrax/commit/6688333c9343fd7d2e19594661522a826d871951

hweng:
@lsitu Sounds good. Just upgrade to the latest commit. We might need to keep on upgrading it in each  sprint.

lsitu:
@mcritchlow I tried to feed Rubocop to make it happy but still got it failed, though I've made lots of unnecessary changes for a shorter variable name, using if block instead of appending the modifier at the end, adding comments like "# frozen_string_literal: true" to all class/module, disabled line length check and method length check in some cases, manipulating for indentation, using the %i pattern as suggested for array etc. But I still can't get it passed for the following cases:
1. "Style/FrozenStringLiteralComment: Missing magic comment # frozen_string_literal: true": when the comment is added on the top of the class definition.
2. "Layout/EmptyLineBetweenDefs: Use empty lines between method definitions.": with method authorize_download! from hyrax that is overridden.
3. "Style/SymbolArray: Use %i or %I for an array of symbols.", while complaining about "Style/PercentLiteralDelimiters: %i-literals should be delimited by [ and ]." when something like %i{one two three} applied.

During feeding Rubocop to fix the errors, there are some inconsistency message like "Layout/TrailingBlankLines: Final newline missing.", "Layout/IndentationConsistency: Inconsistent indentation detected.", "Performance/StringReplacement: Use tr instead of gsub." etc. 

It seems like we should run into the same situation that hyrax do. I think we may need to consider the following changes:
1. Reasonable longer line length, maybe 100 or 120?
2. Allow more lines for methods, maybe 14 to make it inline with hyrax?
3. Use blanket pattern for arrays.
4. Disable FrozenStringLiteralComment to avoid "Missing magic comment # frozen_string_literal: true" error.

mcritchlow:
@lsitu, this should probably be a team decision since everyone is impacted, so @VivianChu and @hweng please give a 👍 or 👎 to each of Longshou's proposed changes to the rubocop config above.

I think it is a very good argument to align with the Hyrax rubocop conventions since we periodically have to override parts of that codebase. My thoughts on each:

1. 👍 I think up to 120 is fine
2. 👍 seems like a reasonable compromise
3. 👍 I don't have strong preferences on this.
4. I could go either way on this. I hadn't heard of this one, to be honest. But it looks like it's trying to give a warning/heads up about the Ruby 3 migration path. Does adding the comment line `# frozen_string_literal: true` as noted in the [cop description](http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/FrozenStringLiteralComment) make this cop happy?

lsitu:
@mcritchlow For #4, the cop just won't be happy with comment `# frozen_string_literal: true` added on the top of the module definition for modules like FileSetIndexer and IndexesAttributes even. But I can give it another try to add it to the body of the module definition to see how it goes.

mcritchlow:
@lsitu if it didn't work, I'd say disable as you proposed. Not worth the hassle then, imo

hweng:
It is fine for me :+1:.  Overall we have to agree that the rubocop is a great tool which points out some bad style in your code to fix.

VivianChu:
👍

lsitu:
@ucsdlib/developers Per our discussions above, I've added PR https://github.com/ucsdlib/horton/pull/114 for rubocop configuration. Please review and comments. Thanks.

lsitu:
@ucsdlib/developers The style check is passed now. Please review and comments ... Thanks.

mcritchlow (commented):
Thanks for addressing the rubocop updates Longshou, I know that was a lot. LGTM as a first implementation! nice job 👍

VivianChu:
👍

hweng (approved):
:+1:

PR #109, opened 2017-05-12
Fixes - #93 Upgrade hyrax.

### LINKED ISSUE #93: Upgrade to latest Hyrax

hweng:
Fixes #93

* Upgraded hyrax to the point: https://github.com/projecthydra-labs/hyrax/commit/562b8ffab8f1770a45a7b231c1e573dc0d2455cf
* Fixed failed test.


app/controllers/hyrax/controller.rb
```diff
@@ -4,6 +4,7 @@ module Hyrax::Controller
   included do
     class_attribute :create_work_presenter_class
     self.create_work_presenter_class = Hyrax::SelectTypeListPresenter
+    self.search_state_class = Hyrax::SearchState
```
lsitu: I suspect that the whole controller.rb file is not necessary and could be deleted now. Could you give it a try?

hweng: But the local controller.rb is not identical to hyrax/controller.rb. It seems there are some local implementation in controller.rb. Are you sure it can be safely removed?

lsitu:
@hweng Those minor changes could be the reason of some historical issues. As long as the rspec test suites pass, then I think it'll be safe to delete the controller.rb file.

hweng:
@lsitu  The test would pass if delete controller.rb.
 Just let you aware that there are some methods in local controller.rb but not in hyrax/controller.rb,
For example:
https://github.com/ucsdlib/horton/blob/develop/app/controllers/hyrax/controller.rb#L16

It would be fine to me to delete it if all of you agree.

lsitu:
@hweng I see that, which could be migrated from CC pilot version in the past. Please feel free to delete the whole file since the rspec test passed.

hweng:
@lsitu The controller.rb is removed.

lsitu:
👍

VivianChu:
👍

VivianChu (approved):

PR #117, opened 2017-06-07
Updated configuration for rubocop and auto corrected violations.

lsitu:
Updated configuration for rubocop to make it inline with hyrax and auto corrected violations.

Added a commit with rubocop auto fixes so that we can see more straight forward regarding what kind of style check violations that we need to fix, which will fix those violations like indentations, spacing etc, empty lines.
Also, following the the original configuration, spec checks are excluded at this time. 


lsitu:
@ucsdlib/developers I think I've got all those rubocop style issues addressed. However hound still reported 2 violations without sending out any messages, which are hard to figure out what's going on.
Could you review and comment? Please check it out and make sure`rubocop` pass in your local environment for future development. Thanks.

mcritchlow (approved):
I think this iooks like a good start. It would be good to make sure we're in agreement that when the excluded files are touched in new tickets, some effort should ideally be made to address the style issues. That said, I think this puts us in a much better place, and new code can/will be ran against rubocop. Thanks @lsitu 👍

lsitu:
@mcritchlow Agree. We should try our best to address those style check issues in the to do list when we touch those files later, though some of them are not easy to fix.

VivianChu:
👍

PR #118, opened 2017-06-09
Feature/remove local pages

### LINKED ISSUE #89: Remove local pages implementation, replace with Hyrax default
### Decision
per @gamontoya - We should try it out (Hyrax out of the box) at first and revisit.

Tasks:

- [ ] Remove local Pages implementation
- [ ] Create entries in the routes file for our existing static pages: about, faq, takedown, search-tips. Hang off the root, if possible, so the about page will be: `https://<application-path>/dc/about`, for example.

### Descriptive summary

Currently, out of the box, Hyrax provides a mechanism for defining new pages. This is how the About, Help, Contact, pages are defined. As well as specific sections on the default homepage. We could potentially choose to use this for our local pages needs, in place of the custom implementation we created.

I think we need input from @gamontoya and others on what we really need from these pages. For example, how often are end users going to really create "new" pages? Has that feature actually been helpful? It would seem that hasn't been used in damspas.

Note: Matt can give Gabriela a demo to show the UI/UX of each.

### Rationale

Pros for local implementation:

- Our pages implementation allow end users to create new pages without developer help
- Our pages implementation adds a WYSIWYG editor for managing content

Pros for Hyrax implementation:
- Already exists, and isn't something we need to maintain. Less code on our end means faster test suites, less technical overhead, etc.
- Very simple to add new pages, just a line in  `routes.rb`

### Related work

#53

VivianChu:
Fixes #89 ; refs #89 

Remove local Pages implementation and create entries in the routes file for our existing static pages.


mcritchlow:
LGTM, but will need rebase once #117 gets merged

mcritchlow:
@VivianChu - Ok, Longshou's PR is merged. Looks like a handful of pages need updating via a rebase, as expected.

VivianChu:
@ucsdlib/developers - This branch has been rebased with the develop branch. Please review, comment, etc...  Thanks

db/migrate/20170606223403_drop_pages.rb
```diff
@@ -0,0 +1,5 @@
+class DropPages < ActiveRecord::Migration[5.0]
```
lsitu: @VivianChu Could we delete 20170322200503_create_pages.rb and 20170403173356_rename_to_slug.rb in folder db/migration instead? It looks like that we no longer need these two files.

lsitu:
@VivianChu It looks good 👍 . Just a minor question regarding the related db migration files for pages.

db/migrate/20170606223403_drop_pages.rb
VivianChu: @lsitu - These 2 files have been removed.  Thanks

lsitu:
👍

hweng:
:+1:

hweng (approved):

PR #129, opened 2017-07-31
Implemented the file use model with access control for file downloads.

### LINKED ISSUE #111: Finalize File Use modeling
### Descriptive summary

We need to reach agreement on our mapping of existing file use properties in DAMS4 to DAMS5. Longshou noted this is the last (major) mapping/modeling task that we've yet to finish.

If a meeting is needed let's get that setup.

### Rationale

Because data models.

### Related work

#46 - dependency

lsitu:
Fixes #111 ; refs #111

Implemented the file use model with access control for file downloads.

* Implemented the file use model to allow ingesting files with other uses into the same fileset as the original file.
* Added support to index technical metadata as JSON format for all files in a FileSet for UI display.
* Implemented access control for to restrict the PreservationMasterFile and the TIFF format files for curator-only access.
* Display additional files like PreservationMasterFile in the FileSet viewer and enabled file download for curators.


app/actors/file_set_attach_files_actor.rb
```diff
@@ -0,0 +1,33 @@
+# Actions are decoupled from controller logic so that they may be called from a controller or a background job.
+class FileSetAttachFilesActor < Hyrax::Actors::FileSetActor
```
mcritchlow, line 2: What change was needed here that required us overriding the (presumably) upstream `FileSetAttachFilesActor`?

app/actors/hyrax/actors/create_with_files_actor.rb
```diff
@@ -0,0 +1,62 @@
+module Hyrax
+  module Actors
+    # Creates a work and attaches files to the work
+    class CreateWithFilesActor < Hyrax::Actors::AbstractActor
```
mcritchlow, line 4: What change was needed here that required us overriding the (presumably) upstream `CreateWithFilesActor` ?

app/actors/hyrax/actors/create_with_remote_files_actor.rb
```diff
@@ -0,0 +1,104 @@
+module Hyrax
+  module Actors
+    # Attaches remote files to the work
+    class CreateWithRemoteFilesActor < Hyrax::Actors::AbstractActor
```
mcritchlow, line 4: What change was needed here that required us overriding the (presumably) upstream `CreateWithRemoteFilesActor`?

app/controllers/hyrax/downloads_controller.rb
```diff
@@ -105,5 +113,28 @@ def icon_path(asset_id, absolute_path = false)
         end
         ::ThumbnailPathService.icon_path(VisibilityService.visibility_value(file_set.rights_override))
       end
+
+      # Retrieve the content for relation
+      # @param [string] relation
+      # return [binary]
+      def relation_content(relation)
+        case relation
+        when 'preservation_master_file'
```
mcritchlow: Could `preservation_master_file` be a constant instead of a plain string? Perhaps available via the new `ExtendedContainedFiles`

app/controllers/hyrax/downloads_controller.rb
```diff
@@ -105,5 +113,28 @@ def icon_path(asset_id, absolute_path = false)
...
+          preservation_master_file = asset.attached_files.base.preservation_master_file
+          return preservation_master_file if preservation_master_file
+        when 'transcript'
```
mcritchlow: Could `transcript` be a constant instead of a plain string? Perhaps available via the new `ExtendedContainedFiles`

app/jobs/attach_files_to_file_set_job.rb
```diff
@@ -0,0 +1,75 @@
+# Converts UploadedFiles into FileSets and attaches them to works.
+class AttachFilesToFileSetJob < AttachFilesToWorkJob
```
mcritchlow, line 2: What change was needed here that required us overriding the (presumably) upstream `AttachFilesToFileSetJob`?

app/models/ability.rb
```diff
@@ -62,4 +66,10 @@ def custom_permissions
     def curation_concerns
       Hyrax.config.curation_concerns
     end
+
+    def download_file_permission
+      can :read, ActiveFedora::File do |file|
+        file.mime_type == 'image/tiff' ? false : true
```
mcritchlow, line 72: Why is this check specifically targeting `image/tiff`? What are we trying to prevent here? The download of source/master files?

app/indexers/file_set_indexer.rb
```diff
@@ -2,4 +2,54 @@
 
 class FileSetIndexer < Hyrax::FileSetIndexer
   self.thumbnail_path_service = ::ThumbnailPathService
+
+  def generate_solr_document
```
mcritchlow, line 6: was there previously no Solr document content generated for FileSets? Or are we overriding what was there and generating our own here?

app/jobs/import_url_with_file_use_job.rb
```diff
@@ -0,0 +1,34 @@
+require 'uri'
+require 'tmpdir'
+require 'browse_everything/retriever'
+
+# Given a FileSet that has an import_url property,
+# download that file and put it into Fedora
+# Called by AttachFilesToWorkJob (when files are uploaded to s3)
+# and CreateWithRemoteFilesActor when files are located in some other service.
+class ImportUrlWithFileUseJob < ::ImportUrlJob
```
mcritchlow, line 9: What change was needed here that required us overriding the (presumably) upstream `ImportUrlWithFileUseJob`?

mcritchlow (commented):
Added some comments/questions. I think my only overall concern is the number of upstream classes that we seem to be overriding. So my questions are largely targeted around trying to understand the large set of overrides.

Perhaps there isn't an alternative, but I admittedly worry a bit about the sustainability/upgrade path. Hoping @VivianChu and @hweng can take a look as well.

This is not to discount the work on this, really great job @lsitu :+1:

app/actors/file_set_attach_files_actor.rb
lsitu: I think we just extend/override the Hyrax::Actors::FileSetActor to accept a relation/file use parameter.

app/actors/hyrax/actors/create_with_files_actor.rb
lsitu: I think we need to override it to extract file uses attributes from the object metadata and extend method attach_files(files, env, file_uses) to accept the file uses parameter for ingest.

app/actors/hyrax/actors/create_with_remote_files_actor.rb
lsitu: The same as above to extract file uses from object metadata and extend it to accept the file uses properties for ingest.

app/controllers/hyrax/downloads_controller.rb
lsitu: Done in commit https://github.com/ucsdlib/horton/pull/129/commits/e8696d3a7ba2dc81566544144ec098610d5cdc12. Thanks.

app/controllers/hyrax/downloads_controller.rb
lsitu: Also done in commit https://github.com/ucsdlib/horton/pull/129/commits/e8696d3a7ba2dc81566544144ec098610d5cdc12. Thanks.

app/jobs/attach_files_to_file_set_job.rb
lsitu: I think we need to extend the functionality to accept the file use parameter and perform ingest files with those files use properties provided as well.

app/models/ability.rb
lsitu: This is implemented because Ryan's comment for not allowing public users to download the TIFF files, see  https://github.com/ucsdlib/horton/issues/111#issuecomment-316482056,.

app/indexers/file_set_indexer.rb
lsitu: I think hyrax does have solr document generate for FileSets but unfortunately it won't support ingest/store multiples files in one fileset at this time. And there are no strategic to store technical metadata for other files other than the original file in the same fileset. So we need to extend it to include technical metadata for all other files in Solr document so that we can expose them for UI displaying.

app/jobs/import_url_with_file_use_job.rb
lsitu: I think the main purpose for overriding here is also to extend the functionality to accept the file_use/relation parameter and ingest the file with the file use specified

app/models/ability.rb
mcritchlow: I see. So is this the primarily place where we're ensuring `original_file`'s aren't downloaded by non-curators?

Because if so, it seems like we have far more file types than `tiff` which would fall under this check. My reading of Ryan's comment is tiff was just a (common) example. Not the sole use case or instance where we have this use case. And, it also seems like it would be nice if we could check the `file_use` instead of the `mime_type` to determine download-ability? Should we clarify with Ryan?

app/indexers/file_set_indexer.rb
mcritchlow: Ok, that makes sense. Thanks for clarifying :+1:

app/models/ability.rb
lsitu: @mcritchlow Yes, this is one and there could be the videos/audios as well. But it will require deeper and tricker overriding in the derivative creative and technical metadata extraction layer if we change the strategic to ingest a primary file with File Use other than the OriginalFile, which will make it harder for us to maintain our codes over time. So it seems like checking for mime_type la cleaner solution for us at the moment.

VivianChu:
👍

mcritchlow:
Noting @lsitu 's [post on samvera google group](https://groups.google.com/forum/#!topic/samvera-tech/9mYsteFRgSY) which finally got some discussion going. I'm going to attend the tech call tomorrow to get more feedback from folks. Then we can circle back and discuss next steps with this. So let's hold the merge for now, please.

PR #130, opened 2017-08-02
Suppress components from search results

### LINKED ISSUE #124: R&D Suppress components from search results
### Descriptive summary

Please look into our options for suppressing components from search results. Using a custom SearchBuilder, or whatever makes most sense.

If a possible solution is found, please post screenshots and we can potentially move towards a PR in this ticket as well.

### Rationale

While we don't yet know exactly what we want the user experience to be, as it relates to components, we do know that we want them suppressed from the search results interface.

### Expected behavior

A user searches for "fish" and a component with that metadata will show in the search results. We only want the top-level object to be returned, not the component.

### Actual behavior

All matching records, including components, are returned in the search results.

VivianChu:
Fixes #124 ; refs #124 

Suppress components from search results

### DIFF
--- a/app/controllers/hyrax/object_resources_controller.rb
+++ b/app/controllers/hyrax/object_resources_controller.rb
@@ -17,7 +17,7 @@ module Hyrax
     def new
       curation_concern.created_date_attributes = [{ label: 'LABEL' }]
       # trick to trigger the form for nested attributes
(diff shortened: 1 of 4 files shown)


mcritchlow (commented):
LGTM :+1:

lsitu (approved):
Nice! 👍

PR #137, opened 2017-09-07
Fixes #136 - Added IIIF support with OpenSeadragon and Cantaloupe.

### LINKED ISSUE #136: IIIF R&D
### Descriptive summary

Initial evaluation of IIIF options, particularly based on the excellent experience written up by Jonathan at CHF: https://bibwild.wordpress.com/2017/08/29/full-res-pan-and-zoom-js-viewer-on-a-sufiahyrax-app/

Of particular interest are:

- Evaluating [Cantaloupe](https://github.com/medusa-project/cantaloupe)
- Evaluating `vips` vs `graphicsmagick` vs `imagemagick` to generate static tilesets (as we've done w/ DAMS4). The performance gains from `vips` as written about, as well as `graphicsmagick` over `imagemagick` seem well worth exploring.

### Rationale

Based on the information we have available, we want to try and make the best decision about what to do with image derivatives.

lsitu:
Fixes #136

Added initial IIIF support with the OpenSeadragon viewer on Cantaloupe IIIF server.

* Added the default OpenSeadragon viewer
* Enabled Cantaloupe to retrieve the master files from Fedora filestore 
* Make IIIF support configurable for each environment.


mcritchlow:
Assuming we're going to go with openseadragon, it may be worth considering this gem which provides a simplified asset packaging and some helpers - https://github.com/cbeer/openseadragon-rails

app/views/hyrax/file_sets/media_display/_image.html.erb
```diff
@@ -0,0 +1,44 @@
+<%
```
mcritchlow: Do we need to do all this here, or is the [representative media partial](https://github.com/samvera/hyrax/blob/master/app/views/hyrax/base/_representative_media.html.erb) sufficient?

app/views/hyrax/file_sets/show.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="container-fluid">
```
mcritchlow: Likewise to the other comment, does this belong in the Fileset view, or the [base show partial](https://github.com/samvera/hyrax/blob/master/app/views/hyrax/base/show.html.erb) for the object itself, with the representative media partial perhaps moved from it's default location to a more prominent place? (that's also something we could sort out later w/ usability study)

mcritchlow (commented):
A few additional comments/questions

app/views/hyrax/file_sets/media_display/_image.html.erb
lsitu: It does seem like we need to override the default behavior of mage display for IIIF. I think we can move the codes for the iiif_url to a helper class instead, otherwise the override of the Hyrax::FileSetsControl for a custom FileSetPresenter with the function to generate the iiif_url for the image. What do you think?

app/views/hyrax/file_sets/show.html.erb
lsitu: Yes, it sounds like something that we can could sort out later. I just override it for the OpenSeadragon viewer to be display properly for now.

app/views/hyrax/file_sets/media_display/_image.html.erb
mcritchlow: But isn't that partial for the display of an image within the context of the fileset section of the view, not the primary representative media for the object show page? I would think the latter is presumably the context in which we'd want to present a IIIF viewer?

Or maybe put another way, if we don't change the representative media partial, are we going to be left with a static image display, but have the IIIF viewer elsewhere, displayed in the FileSet context? That seems like it might be odd.

app/views/hyrax/file_sets/show.html.erb
mcritchlow: Yeah, agreed. The exact placement in the view we can sort out later 👍 I just want to make sure we're housing it in the correct partial.

app/views/hyrax/file_sets/media_display/_image.html.erb
lsitu: Right, it's the FileSet section of the viewer that could serve the purpose of testing the functionality of the IIIF and Cantaloupe. We can added it to the object viewer, but it looks like we need some UI works for the object viewer ... Should we make the changes to the object viewer now or later when we get some inputs?

lsitu: @mcritchlow How about making a similar change to the object viewer as for the fileset now, and we adjust the object viewer later once we get some Inputs?

mcritchlow: Do you mean make the change in both places? If so, I'm not sure we need that for this R&D. I guess for now it's more important to try things out than have it done perfectly the first time, as long as we can revert/replace as needed. Any thoughts on your end @VivianChu ?

Also, @lsitu , what are your thoughts about using [the gem I referenced](https://github.com/ucsdlib/horton/pull/137#issuecomment-327883273)? Might make it easier to use and potentially remove if we decide to go with, say, universal viewer instead in the future?

lsitu:
@mcritchlow I've changed it to use the OpenSeaDragon gem and its helper method for the picture viewer. And I've moved the IIIF viewer to the Object page while leave the static thumbnail on the FileSet and object page by themselves. 
Now the IIIF OpenSeaDragon viewer on the Object page will be triggered by the mouseover event and reset by double click. See commit https://github.com/ucsdlib/horton/pull/137/commits/8e3794aeca6c9b8f55b849e8fa81bdd13f1bf0fb.

app/views/hyrax/file_sets/media_display/_image.html.erb
lsitu: @mcritchlow I think it's done. See my comment earlier, https://github.com/ucsdlib/horton/pull/137#issuecomment-328136108.

mcritchlow:
Yeah I think this looks like a great starting point for evaluation. Thanks @lsitu 

Will let @VivianChu take a peek too and sign off on the PR

VivianChu:
👍

VivianChu (approved):

PR #138, opened 2017-09-08
Fixes 136 - Fixed OpenSeaDragon viewer for objects with no representative id and improved test coverage.

### LINKED ISSUE #136: IIIF R&D

lsitu:
Fixes #136

Fixed the error for OpenSeaDragon viewer for objects with no representative id and added more tests to improv test coverages.

### DIFF
--- a/spec/features/object_resource_spec.rb
+++ b/spec/features/object_resource_spec.rb
@@ -244,6 +244,28 @@ feature 'ObjectResource' do
     end
   end
 
+  context 'IIIF OpenSeaDragon viewer' do
+    let(:user) { create(:admin) }
+    let(:object_with_file) { FactoryGirl.create(:object_with_image_file, title: ["Object - IIIF Viewer"], user: user) }
+
+    before do
+      sign_in user
+    end
+
+    scenario 'has no OpenSeaDragon widget' do
+      visit new_hyrax_object_resource_path
+      fill_in 'Title', with: 'Test ObjectResource - IIIF OpenSeaDragon'
+      click_button 'Save'
+      expect(page).to have_content 'Test ObjectResource - IIIF OpenSeaDragon'
+      expect(page).not_to have_selector "div[class='open_seadragon_widget']"
+    end
+
+    scenario 'has OpenSeaDragon widget' do
+      visit hyrax_object_resource_path object_with_file.id
+      expect(page).to have_selector "div[class='open_seadragon_widget']"
+    end
+  end
+
   context 'a logged in user in editor role' do
     let(:user) { create(:editor) }
 
--- a/app/views/hyrax/base/show.html.erb
+++ b/app/views/hyrax/base/show.html.erb
@@ -8,7 +8,7 @@
     </header>
 
     <!-- Add OpenSeaDragon Viewer -->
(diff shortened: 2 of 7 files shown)


mcritchlow:
LGTM, @VivianChu thoughts?

VivianChu:
@mcritchlow : Looks good 👍

spec/features/object_resource_spec.rb
```diff
@@ -244,6 +244,34 @@
     end
   end
 
+  context 'IIIF OpenSeaDragon viewer' do
+    let(:user) { create(:admin) }
+    let(:object_with_file) { FactoryGirl.create(:object_with_image_file, title: ["Object - IIIF Viewer"], user: user) }
+
+    before do
```
mcritchlow, line 251: why is this needed? or rather, what would be the downside of just setting `iiif_enabled` to `true` in the [test environment](https://github.com/ucsdlib/horton/blob/a8f133a7f31417ece7f71c73c9d5b6433de04870/config/environments/test.rb#L62)? it seems you could do away with this before/after block setup/teardown entirely then.

Related, if we set it to true, then it seems we'd have that value set to `true` for all environments, which makes me wonder what the need for the configuration setting is, and the related checks for whether it's true or not, such as [this](https://github.com/ucsdlib/horton/blob/a8f133a7f31417ece7f71c73c9d5b6433de04870/app/views/hyrax/base/show.html.erb#L11)

mcritchlow (commented):
@lsitu - thanks for adding these additional tests. 

I just one question about the config value for whether IIIF is enabled or not, and what value that brings us. Otherwise this looks great.

spec/features/object_resource_spec.rb
lsitu: @mcritchlow I think the original thought is enable us to hide the viewer for any purpose. But since now we've added the mouse over trigger to have it shown up, so I think we can remove it.
I've updated the PR to remove the configuration: https://github.com/ucsdlib/horton/pull/138/commits/a975b474b0711dd81d4c852a74939e7a871c585f.

mcritchlow: Ah, ok. Yeah that sounds good. Thanks! @VivianChu do you have any other questions or concerns, or should we get this merged?

VivianChu: @mcritchlow - I don't have any other question.  We should get this merged.

VivianChu (approved):

PR #139, opened 2017-09-12
Add authentication and authorization

### LINKED ISSUE #125: Implement authentication and authorization
### Descriptive summary

Implement SSO (authentication) + AD/LDAP (authorization) for Horton. Considerations:

1. Existing damspas strategy using omniauth-shib gem
2. ILL Billing strategy that @hweng created.

### Rationale

Once we have a test environment, we'll want to immediately get this setup to begin testing access control.

### Related work

* Dependency on having a test environment

VivianChu:
Fixes #125  ; refs #125 

Implement SSO (authentication) + AD/LDAP (authorization).

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,3 +1,5 @@
+require 'net/ldap'
+
 class User < ApplicationRecord
   # Connects this user object to Hydra behaviors.
   include Hydra::User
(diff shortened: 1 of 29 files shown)


app/models/concerns/hyrax/user.rb
```diff
@@ -0,0 +1,173 @@
+require 'oauth'
```
mcritchlow: what was added to this class that required the override?

app/models/user.rb
```diff
@@ -1,3 +1,6 @@
+require 'net/ldap'
+require 'fakeldap'
```
mcritchlow: why does the user model need fakeldap?

mcritchlow (commented):
A few comments/questions.

@lsitu - if you have a chance to take a look and review, would be much appreciated.

app/models/concerns/hyrax/user.rb
VivianChu: @mcritchlow - These 2 methods are required to override because we don't use password to login.  The batchimporter tool use them.
https://github.com/ucsdlib/horton/blob/feature/ad_ldap/app/models/concerns/hyrax/user.rb#L143
https://github.com/ucsdlib/horton/blob/feature/ad_ldap/app/models/concerns/hyrax/user.rb#L152

app/models/user.rb
VivianChu: @mcritchlow - The user model doesn't need fakeldap.  I added it in while testing and forgot to remove it.  I'll remove it and re-commit.

app/models/concerns/hyrax/user_override.rb
```diff
@@ -0,0 +1,57 @@
+require 'oauth'
+
+module Hyrax::UserOverride
```
lsitu: @VivianChu I am just curios whether this is a good practice to add a custom class in the Hyrax module or not. I think normally we only add those classes that are overridden from the upstream under the "hyrax" folder.

mcritchlow: @lsitu - This was something I was curious about. Essentially, as to whether there is a cleaner option for overriding modules/classes in Hyrax. It does seem quite significant to have to override/copy an entire module or class to add a single method, or update an existing one. For views/partials I can see it, but for classes and modules it seems like ideally there would be a cleaner object-oriented way to have overrides with a lighter footprint. So this issue [came up recently on the samvera google group](https://groups.google.com/forum/#!topic/samvera-tech/Z5bMJfPXZ1c) and @VivianChu was looking into it the feasibility of it.

app/models/concerns/hyrax/user.rb
lsitu: @VivianChu Could we try to override both methods in models/user.rb instead? I think using something like ` def self.audit_user` and `def self.batch_user` with the same codes should work.

It looks like it could be even better if we can override the class method `def self.find_or_create_system_user(user_key)` instead of those two above in models/user.rb. Could you give it a try? Thanks.

app/models/concerns/hyrax/user_override.rb
lsitu: @mcritchlow Right. This is something that we often may need to spend sometime on for a better/cleaner way. But sometimes it could be challenge with the limitation of the hyrax design patterns that only allow us to do that way but not the other way ...

app/models/concerns/hyrax/user.rb
mcritchlow: @VivianChu - did you have a chance to try @lsitu 's recommendation?

VivianChu:
@mcritchlow @lsitu - I tried Longshou's suggestion before it doesn't work.  I tried again today and still got that error.  I think it's because it failed at this part
https://github.com/ucsdlib/horton/blob/feature/ad_ldap/spec/controllers/batch_import_controller_spec.rb#L58
https://github.com/ucsdlib/horton/blob/develop/app/models/user.rb#L3

 because it called hyrax:user model

 1) BatchImportController Excel XL metadata source #create enqueue an update job is successful
     Failure/Error: log.success!
     
     ActiveModel::UnknownAttributeError:
       unknown attribute 'password' for User.

mcritchlow:
That's odd. Using @lsitu's suggestion, your just overriding the methods defined in the module in the included class.

Is there a stacktrace, or any additional info from byebug? I'm curious where/why it's bypassing the methods you're defining locally.

VivianChu:
@mcritchlow - It related to the structure of the module hyrax:user

Here's the stack trace.  It does call back to hyrax:user class.  It doesn't use these methods in model/user.rb

     # /Users/tchu/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/bundler/gems/hyrax-9f117a32a9fd/app/models/concerns/hyrax/user.rb:163:in `find_or_create_system_user'
     # /Users/tchu/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/bundler/gems/hyrax-9f117a32a9fd/app/models/concerns/hyrax/user.rb:146:in `audit_user'
     # /Users/tchu/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/bundler/gems/hyrax-9f117a32a9fd/app/services/hyrax/message_user_service.rb:27:in `job_user'
     # /Users/tchu/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/bundler/gems/hyrax-9f117a32a9fd/app/services/hyrax/message_user_service.rb:11:in `call'
     # /Users/tchu/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/bundler/gems/hyrax-9f117a32a9fd/config/initializers/hyrax_callbacks.rb:29:in `block in <top (required)>'

lsitu:
@VivianChu I've added a commit https://github.com/ucsdlib/horton/pull/139/commits/1f0bfdd1c1e85af2500ee1ddefd612043ea26eb1 that works in my local environment as we discussed this morning. Could you do a clean checkout to see whether it works for you?

VivianChu:
@lsitu - It works for me.  Thanks.

VivianChu:
@ucsdlib/developers - please review, comment, etc... Thanks

VivianChu:
@ucsdlib/developers - The commits have been merged for easier review.  Thanks

hweng:
:+1:

mcritchlow (approved):
I think this is a great start until we have an actual system to test against. Thanks @VivianChu! And to @lsitu for the override idea.

PR #140, opened 2017-09-13
Devops

orangewolf:
Initial ansible stack for the Horton app. Requires encrypted settings for production deployment not found in repo. See README for deploy instructions.


Gemfile
```diff
@@ -2,8 +2,9 @@ source 'https://rubygems.org'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '~> 5.0.0', '>= 5.0.0.1'
-# Use sqlite3 as the database for Active Record
-gem 'sqlite3'
+# Use #sqlite3 as the database for Active Record
+gem 'sqlite3', group: :development
```
mcritchlow: could this be nested in the development group block?

Gemfile
```diff
@@ -2,8 +2,9 @@ source 'https://rubygems.org'
...
+gem 'pg', group: :production
```
mcritchlow: could this be nested in the production group block?

ops/environments/000_cross_env_vars
```diff
@@ -0,0 +1,139 @@
...
+capistrano_user: deploy
+capistrano_group: deploy
+deploy_target: development
+
+# Database
+db: postgres
+db_adapter: postgresql
+hydra_db: hydra
+hydra_db_user: horton
+hydra_db_pass: penguins;13f1
```
mcritchlow, line 45: Make jinja variables also?

ops/deploy.yml
```diff
@@ -0,0 +1,64 @@
+- name: Get timestamp for repoo
+  hosts: localhost
+  tasks:
+    - shell: 'date +%Y%m%d%H%M%S.%5N'
+      register: current_run_timestamp
+    - set_fact: ansistrano_release_version = "{{ current_run_timestamp.stdout }}"
+
+- name: Permissions and folders
```
mcritchlow, line 8: all 3 "Permissions and folders" tasks are essentially the same with the exception of the path. could `with_items` be used here?

Gemfile
mcritchlow: could we put these in the group blocks?

README.md
```diff
@@ -85,3 +85,16 @@ bin/import-authorities
 ### Create an admin set to start ingest
 ```rake hyrax:default_admin_set:create```
 
+
+
+# Deployment
+Deployment is handled via ansible, allows rollback and can be used to provision new machines
```
mcritchlow, line 91: nothing really notes that the development environment requires docker. Could we add a note specifying that somewhere?

ops/environments/000_cross_env_vars
```diff
@@ -0,0 +1,139 @@
+# Anistrano
```
mcritchlow: spelling ;) `# Ansistrano`

ops/environments/000_cross_env_vars
```diff
@@ -0,0 +1,139 @@
...
+  - config/secrets.yml
+  - config/fedora.yml
+  - config/blacklight.yml
+  - config/redis.yml
+  - config/resque-pool.yml
+  - config/solr.yml
+
+ansistrano_deploy_via: "git"
+#ansistrano_git_repo: https://github.com/ucsdlib/horton.git
+ansistrano_git_repo: https://github.com/notch8/horton.git
```
mcritchlow: This is temporary still just for testing I assume?

ops/environments/000_cross_env_vars
mcritchlow, line 45: these should be jinja variables too?

ops/environments/000_cross_env_vars
```diff
@@ -0,0 +1,139 @@
...
+solr_context: solr
+solr_core: collection1
+zookeeper_hosts: "{{groups['horton_solr']}}"
+
+# SSH Keys
+ssh_keys:
+  - winescout
+  - smcfarlane
+  - orangewolf
+# TODO UCSD users
```
mcritchlow, line 142: Yeah we should talk about this. Once this is up and running cleanly, I wonder how much SSH access will/should be necessary

mcritchlow (commented):
An initial review with a few comments/questions. 

Additionally:

1) Do we need to 'own' all these ansible roles internally within the repo? Can we link to them via galaxy instead of having them in git? Or have there been enough changes done to them that we need to go this route.

2) Note the Hound-CI errors, I'm leaning towards having hound ignore `/ops`, agreed?

Gemfile
orangewolf: yes, totally, we've moved away from block form to line by line, but I should have done this in the existing style.  will update

ops/environments/000_cross_env_vars
orangewolf: correct

ops/environments/000_cross_env_vars
orangewolf: just the people who can deploy need access to the deploy user via ssh (which is what this is about)

ops/environments/000_cross_env_vars
orangewolf: @mcritchlow I'm not sure I understand what you mean here. this file is full of config variable for the various parts.  These db creds specifically are the defaults for the dev environment.  they are overridden by an encrypted config file which has a few prod specific things in it.   Here are the keys from that prod file (values left out)


# Prod items
proxy_env
  http_proxy
  https_proxy


# Database
db
db_adapter
hydra_db
hydra_db_user
hydra_db_pass
hydra_db_host
hydra_db_port
fedora_db
fedora_db_user
fedora_db_pass
fedora_db_host
fedora_db_port

deploy_target

# Postfix
postfix_relayhost
postfix_smtp_sasl_auth_enable
postfix_smtp_sasl_user
postfix_smtp_sasl_password
postfix_send_test_mail_to
postfix_rewrite_sender_address

orangewolf:
I agree completely on (2) I think I've accomplished that but I haven't done a lot of hound-ci so we'll see if that worked.  As far as (1) is concerned there are a few groups of things.  

1) repos we didn't change
AnsibleShipyard.ansible-zookeeper
DavidWittman.redis
carlosbuenosvinos.ansistrano-deploy
carlosbuenosvinos.ansistrano-rollback
evrardjp.keepalived
geerlingguy.nodejs
geerlingguy.repo-epel
openmicroscopy.haproxy
repleo.postfix
rvm_io.ruby

2) stuff from https://github.com/curationexperts/ansible-hydra which may or may not have changes to better support RHEL and clustering
hydra-stack
packages
sidekiq
sufia_dependencies
system_setup
webserver

3) anything custom for this install
deploy - these are specific to this rails app hooks for the deploy
sudoersfix - this is needed the docker set up
ucsd.proxy - this is needed to deal with the proxy settings on the configured live machines

4) Notch8 packages
notch8.ffmpeg  - modified version of geerlingguy.ffmpeg
notch8.ssh - I dont think we have this published anywhere, but we probably should

So the easiest ones to deal with are (3) and (4). (3) are pretty specific to this app and belong in the repo. (4) should be published as part of the cleanup of this work set. 

 (2) we'll work on contributing back to the Curation Experts repo what they will take, including a proposal to break it out in to smaller, more usable pieces.  There is substantial refactoring that should be done here at some point  

(1)  These are the easiest to move out of the repo and just pull down with Ansible Galaxy.  But the issue is that these upstream libraries will change. They will go away, break or have different arguments.  So to keep things really repeatable, I recommend a pattern of clear delineation of the roles. So you may want to hold on to them in the app repo as the versions that they are and they can be refreshed from upstream on purpose as desired. We can do that the way we have at the moment, or we can move them in to git submodules.  I'm pretty open about which way we do that.

ops/environments/000_cross_env_vars
mcritchlow: Ah, yes, I was reading this wrong. Thanks for clarifying :+1:

mcritchlow:
Thanks @orangewolf , this helps a lot.

So in your experience, for (1), is using the [ versioning in ansible galaxy](https://docs.ansible.com/ansible/latest/galaxy.html#version) not as dependable as one might hope? I'm honestly curious, because I have essentially no experience using it in practice like you all do.

re: (2) - that sounds great

re: (3,4) - that also sounds great, and makes sense. Thanks!
- only question about ffmpeg, is @jhriv [setup a role](https://github.com/ucsdlib/ansible-role-ffmpeg) a while back in our org. Do you see any opportunity or value in reconciling differences between those and making them one?

orangewolf:
For the most part they will stay stable.  But it only takes one going away or changing focus to break the build. The roles on galaxy essentially just point at github repos. We find this same problem with pointing at remote github repos in our Gemfile.  Did you notice I had to change the Gemfile to fix loading the Samvera dependencies.  The app was essentially un-deployable do to things outside our control.  In my experience that's not a big deal except at 3 am when there is a critical fix that needs to launch.

I'll look at the ansible role this afternoon to see if we can't get synced up there.

jhriv:
@orangewolf, another change we'd like is to remove the roles/* from this PR, and replaced with a `roles.yml` or the like.

For the roles that are modified from upstream, or not in Ansible Galaxy, we (ucsdlib) can fork, and have the modifications applied, or we can pull directly from notch8's organization's github.

```yaml
- name: rvm_io.ruby
  version: v2.0.0

- name: carlosbuenosvinos.ansistrano-deploy
  version: 2.7.0

- name: carlosbuenosvinos.ansistrano-rollback
  version: 2.0.1

- name: notch8.ssh
  src: https://github.com/notch8/ansible-role-ssh.git
  version: master
```

This way we can take advantage of roles in other projects without having to duplicate code.

bin/provision would have to be updated with a call to ansible-galaxy:
`system "ansible-galaxy install --roles-path ops/roles --role-file ops/roles.yml"`

PR #151, opened 2018-04-05
Add pg gem to a production group

mcritchlow:
This seems to have gotten lost in what we received from Notch8. Adding
back in for reference.


jhriv (approved):
The fips.log in gitignore is not a database

